PR #1, opened 2023-01-17
Add Resource binding test

guusdk:
A new test that asserts that the server closes the connection with an error, when a stanza is sent prior to resource binding (as defined by RFC 3620, section 7.1).

Smack doesn't offer an API that can be utilized to send a stanza between authentication and resource binding. For that reason, this test sends stanzas prior to authentiation (instead of post auth, pre bind).


guusdk:
This tests for a bug that is fixed in Openfire 4.8.0: https://igniterealtime.atlassian.net/jira/software/c/projects/OF/issues/OF-2565

This PR depends on the following (or similar) changes in Smack:
- https://github.com/igniterealtime/Smack/pull/542 (to make sure that the test is executed on all connection types, instead of repeated on the same connection)
- https://github.com/igniterealtime/Smack/pull/543 (to allow this test to run against Openfire that uses its default configuration for the port to be used for websockets)

guusdk:
Rebased.

The current build failure is caused by Smack's SNAPSHOT artifacts in maven-central to be compiled with Java 11, which is not backwards compatible with Java 8.

@Flowdalic (assuming that the required changes won't be ported to Smack 4.4.x) will Java 8-support be dropped in Smack 4.5, or is the uploaded artifact incorrect?

guusdk:
As Smack's integration-test module now uses Java 11, this project should follow. I've created https://github.com/XMPP-Interop-Testing/smack-sint-server-extensions/pull/11 to upgrade from Java 8 to 11.

This PR should be rebased after #11 gets merged.

guusdk:
#11 is merged. Rebased this PR onto main.

src/main/java/org/jivesoftware/smack/ResourceBindingIntegrationTest.java
```diff
@@ -0,0 +1,215 @@
...
+import java.util.concurrent.atomic.AtomicReference;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+/**
+ * Tests that verify the behavior of Openfire around resource binding.
+ *
+ * @author Guus der Kinderen, guus.der.kinderen@gmail.com
+ */
+@SpecificationReference(document = "RFC 6120")
```
Fishbowler, line 44:
```suggestion
@SpecificationReference(document = "RFC6120")
```
Remove space to match the junit `specifications.properties`

Fishbowler (approved):

src/main/java/org/jivesoftware/smack/ResourceBindingIntegrationTest.java
Fishbowler: ...or we can change the junit references

Fishbowler: ...or we can change the junit references

PR #4, opened 2024-01-18
XEP-0133 Ad-hoc Command tests

Fishbowler:
For testing the Ad-hoc Commands in Openfire, many of which were added or edited in https://github.com/igniterealtime/Openfire/pull/2381

### DIFF
--- a/src/main/java/org/igniterealtime/smack/inttest/xep0133/AdHocCommandIntegrationTest.java
+++ b/src/main/java/org/igniterealtime/smack/inttest/xep0133/AdHocCommandIntegrationTest.java
@@ -175,17 +175,6 @@ public class AdHocCommandIntegrationTest extends AbstractAdHocCommandIntegration
     }
 
     @SmackIntegrationTest(section = "4.1")
-    public void testAddUserWithoutJid() throws Exception {
-        checkServerSupportCommand(ADD_A_USER);
-        Exception e = assertThrows(IllegalStateException.class, () ->
-            executeCommandWithArgs(ADD_A_USER, adminConnection.getUser().asEntityBareJid(),
-                "password", "password",
-                "password-verify", "password"
-        ));
-        assertEquals("Not all required fields filled. Missing: [accountjid]", e.getMessage());
-    }
-
-    @SmackIntegrationTest(section = "4.1")
     public void testAddUserWithMismatchedPassword() throws Exception {
         checkServerSupportCommand(ADD_A_USER);
         // Setup test fixture.
--- a/src/main/java/org/igniterealtime/smack/inttest/xep0133/AdHocCommandIntegrationTest.java
+++ b/src/main/java/org/igniterealtime/smack/inttest/xep0133/AdHocCommandIntegrationTest.java
@@ -130,8 +130,8 @@ public class AdHocCommandIntegrationTest extends AbstractAdHocCommandIntegration
 
     @SmackIntegrationTest(section = "3", quote =
         "A server or component MUST advertise any administrative commands it supports via Service Discovery (XEP-0030) " +
-            "(as described in XEP-0050: Ad-Hoc Commands); such commands exist as well-defined discovery nodes associated " +
-            "with the service in question.")
+        "(as described in XEP-0050: Ad-Hoc Commands); such commands exist as well-defined discovery nodes associated " +
+        "with the service in question.")
     public void testGetCommandsForAdmin() throws Exception {
         // Setup test fixture.
 
@@ -144,7 +144,7 @@ public class AdHocCommandIntegrationTest extends AbstractAdHocCommandIntegration
     }
 
     //node="http://jabber.org/protocol/admin#add-user" name="Add a User"
-    @SmackIntegrationTest(section = "4.1")
+    @SmackIntegrationTest(section = "4.1", quote = "Adding a user MUST result in the creation of an account [...] The command node for this use case SHOULD be \"http://jabber.org/protocol/admin#add-user\".")
     public void testAddUser() throws Exception {
         checkServerSupportCommand(ADD_A_USER);
         // Setup test fixture.
@@ -174,7 +174,7 @@ public class AdHocCommandIntegrationTest extends AbstractAdHocCommandIntegration
         }
     }
 
-    @SmackIntegrationTest(section = "4.1")
+    @SmackIntegrationTest(section = "4.1", quote = "Adding a user MUST result in the creation of an account [...] The command node for this use case SHOULD be \"http://jabber.org/protocol/admin#add-user\".")
     public void testAddUserWithMismatchedPassword() throws Exception {
         checkServerSupportCommand(ADD_A_USER);
         // Setup test fixture.
@@ -198,7 +198,7 @@ public class AdHocCommandIntegrationTest extends AbstractAdHocCommandIntegration
         }
     }
 
-    @SmackIntegrationTest(section = "4.1")
+    @SmackIntegrationTest(section = "4.1", quote = "Adding a user MUST result in the creation of an account [...] The command node for this use case SHOULD be \"http://jabber.org/protocol/admin#add-user\".")
     public void testAddUserWithRemoteJid() throws Exception {
         checkServerSupportCommand(ADD_A_USER);
         // Setup test fixture.
@@ -219,7 +219,7 @@ public class AdHocCommandIntegrationTest extends AbstractAdHocCommandIntegration
         }
     }
 
-    @SmackIntegrationTest(section = "4.1")
+    @SmackIntegrationTest(section = "4.1", quote = "Adding a user MUST result in the creation of an account [...] The command node for this use case SHOULD be \"http://jabber.org/protocol/admin#add-user\".")
     public void testAddUserWithInvalidJid() throws Exception {
         checkServerSupportCommand(ADD_A_USER);
         // Setup test fixture.
@@ -241,7 +241,7 @@ public class AdHocCommandIntegrationTest extends AbstractAdHocCommandIntegration
     }
 
     //node="http://jabber.org/protocol/admin#delete-user" name="Delete a User"
-    @SmackIntegrationTest(section = "4.2")
+    @SmackIntegrationTest(section = "4.2", quote = "An administrator may need to permanently delete a user account. [...] The command node for this use case SHOULD be \"http://jabber.org/protocol/admin#delete-user\".")
     public void testDeleteUser() throws Exception {
         checkServerSupportCommand(DELETE_A_USER);
         // Setup test fixture.
@@ -262,7 +262,7 @@ public class AdHocCommandIntegrationTest extends AbstractAdHocCommandIntegration
         }
     }
 
-    @SmackIntegrationTest(section = "4.2")
+    @SmackIntegrationTest(section = "4.2", quote = "An administrator may need to permanently delete a user account. [...] The command node for this use case SHOULD be \"http://jabber.org/protocol/admin#delete-user\".")
     public void testDeleteUserWithFullJid() throws Exception {
         checkServerSupportCommand(DELETE_A_USER);
         // Setup test fixture.
@@ -287,7 +287,7 @@ public class AdHocCommandIntegrationTest extends AbstractAdHocCommandIntegration
     }
 
     //node="http://jabber.org/protocol/admin#disable-user" name="Disable a User"
-    @SmackIntegrationTest(section = "4.3")
+    @SmackIntegrationTest(section = "4.3", quote = "An administrator may need to temporarily disable a user account. Disabling a user MUST result in the termination of any active sessions for the user and in the prevention of further user logins until the account is re-enabled")
     public void testDisableUser() throws Exception {
         checkServerSupportCommand(DISABLE_A_USER);
         // Setup test fixture.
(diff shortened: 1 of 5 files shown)


src/main/java/org/jivesoftware/smackx/commands/AdHocCommandIntegrationTest.java
```diff
@@ -0,0 +1,1425 @@
...
+
+            // Execute system under test.
+            AdHocCommandData result = executeCommandSimple(GET_NUMBER_OF_DISABLED_USERS, adminConnection.getUser().asEntityBareJid());
+
+            // Verify results.
+            assertTrue(Integer.parseInt(result.getForm().getField("disabledusersnum").getFirstValue()) >= 1);
+        } finally {
+            // Tear down test fixture.
+            deleteUser(disabledUser);
+            // TODO consider unmarking the user as being disabled, as deleting the user might not propagate.
```
Fishbowler, line 1012: @guusdk Wouldn't this be a bug in the server implementation?

guusdk: Yes, I think so.

guusdk:
Rebased

guusdk:
I've added the new Smack `@SpecificationReference` annotation (and friends). This makes it obvious that the current implementation tests both XEP-0133-specified commands, but also Openfire-specific ones. Maybe split those up over two distinct files?

guusdk:
I've noticed that these tests uses assertions from this source: `org.junit.Assert.*` - Most (all?) other tests use `org.junit.jupiter.api.Assertions.*` instead.

I'm not sure if there's a functional difference. One difference is the order of arguments being used in the various assertions. For consistency, lets use the same assertions everywhere.

guusdk:
Rebased.

src/main/java/org/jivesoftware/smackx/commands/AdHocCommandIntegrationTest.java
```diff
@@ -0,0 +1,1477 @@
...
+import java.time.ZonedDateTime;
+import java.time.format.DateTimeFormatter;
+import java.util.*;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import static org.junit.Assert.*;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+@SpecificationReference(document = "XEP-0133")
```
Fishbowler, line 42: Needs a version

Fishbowler:
Rebased

Fishbowler:
Still to do:
 * Tests for s4.22
 * Tests for s4.24
 * Tests for s4.25
 * Tests for s4.26
 * Tests for s4.27
 * Tests for s4.28
 * Tests for s4.30
 * Tests for s4.31
 * Explicit `fail(...)` messaging for better output

pom.xml
```diff
@@ -86,6 +86,10 @@
                             <key>sinttest.securityMode</key>
                             <value>disabled</value>
                         </systemProperty>
+                        <systemProperty>
```
guusdk, line 89: Is this an artifact of your local development that wasn't supposed to be committed, or is this an intentional change?

pom.xml
```diff
@@ -163,5 +167,10 @@
             <url>https://oss.sonatype.org/content/repositories/snapshots</url>
             <snapshots><enabled>true</enabled></snapshots>
         </repository>
+
+        <repository>
```
guusdk, line 171: I don't think a new dependency was added in this PR. Why the new repository definition?

src/main/java/org/jivesoftware/smackx/commands/AdHocCommandVendorSpecificIntegrationTest.java
```diff
@@ -0,0 +1,581 @@
...
+    private static final String GET_LIST_OF_GROUP_MEMBERS = "http://jabber.org/protocol/admin#get-group-members";
+    private static final String GET_LIST_OF_EXISTING_GROUPS = "http://jabber.org/protocol/admin#get-groups";
+    private static final String GET_BASIC_STATISTICS_OF_THE_SERVER = "http://jabber.org/protocol/admin#get-server-stats";
+    private static final String GET_NUMBER_OF_CONNECTED_USER_SESSIONS = "http://jabber.org/protocol/admin#get-sessions-num";
+    private static final String GET_USER_PROPERTIES = "http://jabber.org/protocol/admin#get-user-properties";
+    private static final String CURRENT_HTTP_BIND_STATUS = "http://jabber.org/protocol/admin#status-http-bind";
+    private static final String UPDATE_GROUP_CONFIGURATION = "http://jabber.org/protocol/admin#update-group";
+    private static final String REQUEST_PONG_FROM_SERVER = "ping";
+
+    public AdHocCommandVendorSpecificIntegrationTest(SmackIntegrationTestEnvironment environment)
```
guusdk, line 59: Should this throw a `TestNotPossible` based on the result of a `jabber:iq:version` check, to see what the make/model of the server under test is?

pom.xml
Fishbowler: This was when the branch needed to grab a Smack snapshot, which no longer appears in the diff since the last rebase. Will revert this.

pom.xml
Fishbowler: Correct. This was a commit to add a simple exec:java config. Will revert.

src/main/java/org/jivesoftware/smackx/commands/AdHocCommandVendorSpecificIntegrationTest.java
Fishbowler: Good idea!

Fishbowler:
This is a good chunk closer to Done. 

There's some errors I'd still like to tidy up in the Openfire vendor-specific tests.

Fishbowler:
@guusdk I think I've fixed all the things you rightly pointed out.

src/main/java/org/igniterealtime/smack/inttest/xep0133/AdHocCommandIntegrationTest.java
```diff
@@ -0,0 +1,1049 @@
...
+            "accountjids", disabledUser.toString()
+        );
+
+        // Verify results.
+        assertNoteType(AdHocCommandNote.Type.error, result);
+        assertNoteContains("Cannot re-enable remote user: " + disabledUser, result);
+    }
+
+    //node="http://jabber.org/protocol/admin#end-user-session" name="End User Session"
+    @SmackIntegrationTest(section = "4.5")
```
guusdk, line 402: It would be best all tests to have a 'quote' (that describes what specific definition is subject of the test), and for all assertions to have a message (that is shown when the assertion fails). This is an example of a failing test that doesn't have such a quote, and not a message in the failed assertion either:

```
💀 The following 1 tests failed! 💀
• XEP-0133: Service Administration, Section 4.5
      "null"
  Failure reason  : expected: <true> but was: <false>
  Stanza log file : ./target/logs/AdHocCommandIntegrationTest.testEndUserSession (Normal).log
  Test class      : class org.igniterealtime.smack.inttest.xep0133.AdHocCommandIntegrationTest
  Test method     : testEndUserSession
  ```
  
  It's next to impossible for anyone to explain why this test is failing.

File: src/main/java/org/igniterealtime/smack/inttest/xep0133/AdHocCommandIntegrationTest.java [hunk 7]
```diff
@@ -0,0 +1,1049 @@
...
+    public void testEndUserSession() throws Exception {
```
guusdk, line 403: The XEP reads:

> If the JID is of the form <user@host>, the service MUST end all of the user's sessions; if the JID is of the form <user@host/resource>, the service MUST end only the session associated with that resource.

Shouldn't this be _two_ tests, instead of one (to match each of the MUST definitions)?

src/main/java/org/igniterealtime/smack/inttest/xep0133/AdHocCommandIntegrationTest.java
```diff
@@ -0,0 +1,1049 @@
...
+            // Login as the user to be able to end their session
+            AbstractXMPPConnection userConnection = environment.connectionManager.getDefaultConnectionDescriptor().construct(sinttestConfiguration);
+            userConnection.connect();
+            userConnection.login(userToEndSession.getLocalpartOrThrow().toString(), "password");
+
+            AdHocCommandData result = executeCommandWithArgs(GET_LIST_OF_ACTIVE_USERS, adminConnection.getUser().asEntityBareJid(),
+                "max_items", "25"
+            );
+            List<String> jids = result.getForm().getField("activeuserjids").getValues().stream().map(CharSequence::toString).collect(Collectors.toList());
+            assertTrue(jids.contains(userConnection.getUser().asEntityBareJidString()));
```
guusdk, line 427: I don't like using an assertion to check if the test fixture has been set up correctly. If the assertion fails, it suggests that the feature-under-test is failing. That's not the case here. Maybe a `throws new TestNotPossible()` (with a descriptive message) is more appropriate.

Additionally, using GET_LIST_OF_ACTIVE_USERS for this purpose puts a strict requirement that its implementation is giving a truly 'live' representation. I can image that this isn't always the case. It would be better to somehow wait (with a timeout) for the user to become part of this list.

Do we need to use this list at all? We only need to detect that the user is kicked, right? Can't we determine that based on some kind of event listener on the connection that we're kicking?

src/main/java/org/igniterealtime/smack/inttest/xep0133/AdHocCommandIntegrationTest.java
```diff
@@ -0,0 +1,1049 @@
...
+
+            // End the user's session
+            result = executeCommandWithArgs(END_USER_SESSION, adminConnection.getUser().asEntityBareJid(),
+                "accountjids", userToEndSession.toString()
+            );
+
+            assertNoteType(AdHocCommandNote.Type.info, result);
+            assertNoteContains("Operation finished successfully", result);
+
+            result = executeCommandWithArgs(GET_LIST_OF_ACTIVE_USERS, adminConnection.getUser().asEntityBareJid(),
```
guusdk, line 437: I would prefer not using another adhoc command to assert that this test is correct. Can't we simply look at the connection state of `userConnection`?

To avoid race conditions, it probably is best to introduce some kind of event listener that waits for a short time to see if the connection is killed (instead of depending on the fact that the very first snapshot of the list of active users is immediately up-to-date).

guusdk (changes requested):
I have reviewed one test in detail: `testEndUserSession`. To avoid a lot of duplication, let's first see if we can apply the review comments to other tests.

src/main/java/org/igniterealtime/smack/inttest/xep0133/AdHocCommandIntegrationTest.java
```diff
@@ -0,0 +1,1049 @@
...
+            );
+            List<String> jids = result.getForm().getField("activeuserjids").getValues().stream().map(CharSequence::toString).collect(Collectors.toList());
+            assertTrue(jids.contains(userConnection.getUser().asEntityBareJidString()));
+
+            // End the user's session
+            result = executeCommandWithArgs(END_USER_SESSION, adminConnection.getUser().asEntityBareJid(),
+                "accountjids", userToEndSession.toString()
+            );
+
+            assertNoteType(AdHocCommandNote.Type.info, result);
```
guusdk, line 434: Should we assert the 'note' element? It is not present in any of the [XEP-0133](https://xmpp.org/extensions/xep-0133.html) examples (or mentioned anywhere in its text). My reading of [XEP-0050](https://xmpp.org/extensions/xep-0050.html) is that they're optional. If they're optional, we probably shouldn't test them. If we do test them, then the assertion should have a message.

guusdk:
I have added [a commit that addresses much of my review points](https://github.com/XMPP-Interop-Testing/smack-sint-server-extensions/pull/4/commits/9d84f58f528de2880ea41e4b4435586ceb06b26d) for the End User Session command. Let's discuss if this is the right approach, after which I can apply similar changes to other commands.

src/main/java/org/igniterealtime/smack/inttest/xep0133/AdHocCommandIntegrationTest.java
guusdk: I have committed changes that address this in all tests that are part of this PR.

src/main/java/org/igniterealtime/smack/inttest/xep0133/AdHocCommandIntegrationTest.java
guusdk: I have committed additional tests.

src/main/java/org/igniterealtime/smack/inttest/xep0133/AdHocCommandIntegrationTest.java
guusdk: I have committed a change where there are two new assertions, assertCommandCompletedSuccessfully and assertCommandFailed. They no longer solely depend on there being a 'note'.

src/main/java/org/igniterealtime/smack/inttest/xep0133/AdHocCommandIntegrationTest.java
guusdk: Where possible, I switched using adhoc commands to assert a certain state with something that uses an event listener.

guusdk (approved):
I believe that I have now addressed all of the issues that I raised earlier. This amounted to more work than anticipated. It would be good for @Fishbowler to review my work again, given the large amount of changes.

Fishbowler:
Thanks Guus! I've been through all of these changes, commit by commit, and they're all ace. There was a lot of cruft from when this was intended for Openfire only that I was blind to.

PR #10, opened 2024-03-09
Add Github Action

Fishbowler:
Add build and test to Github Actions.

The build is matrix, to check compatibility with other Java versions

The test might fail because either tests are broken, or the server under test doesn't pass the tests that are provided. Humans will need to review a test failure to understand.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,6 +2,10 @@ name: Build & Test
 
 on: [pull_request]
 
+env:
+  OPENFIRE_VERSION: 4_8_1 # The version of Openfire to use, using underscores, to match artifact filenames in https://github.com/igniterealtime/Openfire/releases/
+
+
 jobs:
   build:
     name: Build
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -0,0 +1,77 @@
...
+    runs-on: ubuntu-latest
+    needs: build
+    steps:
+      - name: Checkout Openfire actions e.g. 'startCIServer'
+        uses: actions/checkout@v4
+        with:
+          repository: igniterealtime/Openfire
+          sparse-checkout: |
+            .github
+      - name: Download and untar Openfire
```
guusdk, line 40: Can we somehow cache this download? That'd save us quite a few bytes - and possibly speed things up too.

Fishbowler: Done. Seemingly Github was and is running the download OR cache restore, then the untar all in ~1s, but this definitely feels better.

PR #15, opened 2024-04-10
Add ability to store XMPP logs in files

guusdk:
This adds a new Smack debug logger that can be used to store XMPP traffic in distinct files per test.

To configure the logger, use these system properties when invoking the tests:

```
-Dsinttest.debugger="org.igniterealtime.smack.inttest.util.FileLoggerFactory"
-DlogDir=target/logs
```

The `logDir` property value is used to identify the directory in which the logs will be stored. The directory will be created if it doesn't already exist.

### DIFF
--- a/src/main/java/org/igniterealtime/smack/inttest/util/FileLogger.java
+++ b/src/main/java/org/igniterealtime/smack/inttest/util/FileLogger.java
@@ -50,7 +50,7 @@ public class FileLogger extends AbstractDebugger
         if (testUnderExecution != null) {
             filename = testUnderExecution.toString();
         } else {
-            filename = "output";
+            filename = "test_suite_orchestration";
         }
         filename += ".log";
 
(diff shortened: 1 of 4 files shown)


guusdk:
Using the new debug logger, the GitHub workflow that runs all checks in this repository will now expose the XMPP debug logs as an artifact of the GitHub Action.

src/main/java/org/igniterealtime/smack/inttest/util/FileLogger.java
```diff
@@ -0,0 +1,79 @@
...
+        synchronized (dateFormatter) {
+            formatedDate = dateFormatter.format(new Date());
+        }
+
+        String filename;
+        final SmackIntegrationTestFramework.ConcreteTest testUnderExecution = SmackIntegrationTestFramework.getTestUnderExecution();
+        if (testUnderExecution != null) {
+            filename = testUnderExecution.toString();
+        } else {
+            filename = "output";
```
Fishbowler, line 53: Minor, but I'd like to change this to something more obvious, given it's the log for test framework setup and teardown.

```suggestion
            filename = "test_suite_orchestration";
```

Fishbowler (changes requested):
One minor change, but otherwise, this is _awesome_
Observability++

Fishbowler (approved):

PR #16, opened 2024-04-11
Release 1.1

guusdk:
This includes a set of commits, one of which should be the v1.1 release.

In preparation of the release, the version of Smack is downgraded to a non-SNAPSHOT release. This allows us to have most new features of Smack, but still use a stable release.

I've also included a commit that applied the reference-tags to the existing tests in this repo (which _strictly speaking_ could have gone in another PR).


Fishbowler (approved):
So much s/nodename/nodeName 😂

PR #19, opened 2024-04-19
De-duplicate PubSubIntegrationTest

guusdk:
Both this project and the Smack project define the class `org.jivesoftware.smackx.pubsub.PubSubIntegrationTest`. It has been observed that as a result, re-packaging the project can overwrite one class with the other.

To prevent that issue, the class in this project has been renamed. Duplicate tests have been removed.


Flowdalic:
Did you consider moving the class into another package? It is often not ideal if a Module re-uses a package that is already declared by another module. I placed the test in the same package as the API under test to be able to access package-protected API. Not sure if you have the same requirement in server-sinttest.

guusdk:
The same problem persists. I originally copied it in a different package, which immediately caused a compilation issue. I didn't investigate further.

PR #20, opened 2024-04-26
Add tests for XEP-0030 Service Discovery

guusdk:


guusdk:
Note that these tests uncovered missing support in Openfire. That is documented in OF-2825 and is addressed in https://github.com/igniterealtime/Openfire/pull/2460

guusdk:
Rebased.

Fishbowler:
Rebased some more :)

Fishbowler:
Are these failures legit? Or just in the pipeline? :thinking:

guusdk:
The failures for the tests for XEP-0045 section 7 seem to be flappy. At least some failures for XEP-0030 are probably 'valid' - they should be fixed with https://github.com/igniterealtime/Openfire/pull/2460

The flappyness of XEP-0045 is not something to be addressed in this particular PR. As for the XEP-0030 ones: lets merge this, then use the resulting test suite to test https://github.com/igniterealtime/Openfire/pull/2460 :)

PR #22, opened 2024-05-09
Improve JUnitXmlTestRunResultProcessor's URL generation

### LINKED ISSUE #21: Improper characters being used to construct URL in JUnitXmlTestRunResultProcessor
This was logged by continuous integration:

```
WARNING: Invocation of TestRunResultProcessor org.igniterealtime.smack.inttest.util.JUnitXmlTestRunResultProcessor@6562cc23 failed.
java.lang.IllegalArgumentException: Illegal character in fragment at index 45: https://xmpp.org/extensions/xep-0045.html#7.1 & 7.2.2
	at java.base/java.net.URI.create(URI.java:883)
	at org.igniterealtime.smack.inttest.util.JUnitXmlTestRunResultProcessor.process(JUnitXmlTestRunResultProcessor.java:215)
	at org.igniterealtime.smack.inttest.SmackIntegrationTestFramework.main(SmackIntegrationTestFramework.java:119)
Caused by: java.net.URISyntaxException: Illegal character in fragment at index 45: https://xmpp.org/extensions/xep-0045.html#7.1 & 7.2.2
	at java.base/java.net.URI$Parser.fail(URI.java:2913)
	at java.base/java.net.URI$Parser.checkChars(URI.java:3084)
	at java.base/java.net.URI$Parser.parse(URI.java:3128)
	at java.base/java.net.URI.<init>(URI.java:600)
	at java.base/java.net.URI.create(URI.java:881)
	... 2 more
```
It appears that some content that's used to generate a URL is not properly escaped.

guusdk:
The processor that generates JUnit-compatible XML reports attempts to generate an URL to the specifications that are under test.

Input used for these URLs should be properly escaped.

Links should be generated not only for XEP-specifications, but also for RFCs.

If a link can't be created, prevent the report generation from crashing. Instead, just generate a report without a link.

fixes #21

### DIFF
--- a/src/main/java/org/igniterealtime/smack/inttest/util/JUnitXmlTestRunResultProcessor.java
+++ b/src/main/java/org/igniterealtime/smack/inttest/util/JUnitXmlTestRunResultProcessor.java
@@ -347,7 +347,7 @@ public class JUnitXmlTestRunResultProcessor implements SmackIntegrationTestFrame
         if (specification == null || specification.isBlank()) {
             return "";
         }
-        return specification.replaceAll("\\s", "").toUpperCase();
+        return specification.replaceAll("[\\s-]", "").toUpperCase();
     }
 
     public static Duration getAggregatedTime(final SmackIntegrationTestFramework.TestRunResult testRunResult) {
--- a/src/main/java/org/igniterealtime/smack/inttest/util/JUnitXmlTestRunResultProcessor.java
+++ b/src/main/java/org/igniterealtime/smack/inttest/util/JUnitXmlTestRunResultProcessor.java
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2024 Guus der Kinderen
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package org.igniterealtime.smack.inttest.util;
 
 import org.igniterealtime.smack.inttest.FailedTest;
@@ -22,6 +37,8 @@ import java.io.IOException;
 import java.io.OutputStream;
 import java.lang.reflect.Method;
 import java.net.URI;
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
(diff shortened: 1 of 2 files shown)


src/main/java/org/igniterealtime/smack/inttest/util/JUnitXmlTestRunResultProcessor.java
```diff
@@ -239,6 +252,53 @@ public void process(SmackIntegrationTestFramework.TestRunResult testRunResult)
...
     }
 
+    public static URI generateLink(final String specification, final String specificationSection) {
+        if (specification == null || specification.isBlank()) {
+            return null;
+        }
+
+        String link;
+        if (specification.toUpperCase().startsWith("XEP")) {
+            String normalizedSpec = specification.substring(0, 3);
```
Fishbowler, line 262: You want to preserve the author's casing here?

src/test/java/org/igniterealtime/smack/inttest/util/JUnitXmlTestRunResultProcessorTest.java
```diff
@@ -0,0 +1,227 @@
...
+        final String specification = applySmackNormalization("xep-0030");
+        final String specificationSection = null;
+
+        // Execute system under test.
+        final URI result = JUnitXmlTestRunResultProcessor.generateLink(specification, specificationSection);
+
+        // Verify results.
+        assertNotNull(result);
+        assertEquals("https://xmpp.org/extensions/xep-0030.html", result.toString());
+    }
```
Fishbowler, line 133: Is this the same as testXepWithoutSection?

Fishbowler (commented):
I've added a couple of comments.

I'm not sure that these tests prove the right thing. They prove the method works when its been through that normalisation method, but is that true in the normal execution? My tracing couldn't find anywhere it's used outside of the collection at the end.

src/main/java/org/igniterealtime/smack/inttest/util/JUnitXmlTestRunResultProcessor.java
guusdk: Yes, I do. I'm expected smack-normalized input (although I don't know if that's lower or uppercase).

src/test/java/org/igniterealtime/smack/inttest/util/JUnitXmlTestRunResultProcessorTest.java
guusdk: Hehe, possibly. It's me going through all permutations in my head.

Still, there might be value to keep this. If we ever end up modifying its duplicate for reasons© then it's good to have this explicit test that uses a dash, which is documented to be an integral part of the test by the method name. Execution-wise, the duplication costs next to nothing.

guusdk:
> They prove the method works when its been through that normalisation method, but is that true in the normal execution?

You might be right. I assumed it was, but on a closer look, I believe that Smack only normalises when it applies the enabled/disabled-specification configuration.

I'm thinking that it would be kind of nice if Smack's output would actually contain normalized references, as that would potentially prevent a lot of follow-up issues. I'll see if I can add that to Smack.

Fishbowler (approved):

PR #23, opened 2024-05-14
Add tests for XEP-0352: Client State Indication

guusdk:


src/main/java/org/igniterealtime/smack/inttest/xep0352/CsiIntegrationTest.java
```diff
@@ -0,0 +1,155 @@
...
+            throw new TestNotPossibleException("Domain does not seem support XEP-0352 Client State Indication.");
+        }
+    }
+
+    @SmackIntegrationTest(section = "4.2", quote = "If a client wishes to inform the server that it has become inactive, it sends an <inactive/> element in the 'urn:xmpp:csi:0' namespace")
+    public void markActiveTest() throws SmackException.NotConnectedException, InterruptedException
+    {
+        ClientStateIndicationManager.inactive(conOne);
+        // TODO do we need to reset the CSI-state of conOne to not affect other tests?
+    }
```
Fishbowler, line 55: I see the problem you mentioned. This test relies on the Smack implementation erroring noisily if the server does something unexpected as its only version of assertion.

src/main/java/org/igniterealtime/smack/inttest/xep0352/CsiIntegrationTest.java
```diff
@@ -0,0 +1,155 @@
...
+        ClientStateIndicationManager.inactive(conOne);
+        ClientStateIndicationManager.active(conOne);
+        // TODO do we need to reset the CSI-state of conOne to not affect other tests?
+    }
+
+    @SmackIntegrationTest(section = "6", quote = "To protect the privacy of users, servers MUST NOT reveal the clients active/inactive state to other entities on the network.")
+    public void detectDiscoInfoChangeTestUnsubscribed() throws SmackException.NotConnectedException, InterruptedException, XMPPException.XMPPErrorException, SmackException.NoResponseException, SmackException.NotLoggedInException, TestNotPossibleException
+    {
+        IntegrationTestRosterUtil.ensureBothAccountsAreNotInEachOthersRoster(conOne, conTwo);
+        detectDiscoInfoChangeTest();
```
Fishbowler, line 75: I love this variation

Fishbowler (approved):
Seems reasonable.
I don't think, given the tests that are here, that you need to worry about resets. If these tests flake, it'd highlight a bug in the implementation of the server that a client can't repeatedly report itself as active or inactive, and that's allowed.

It'd be great if there were a way to prove that there were no response to the CSI nonza, but given parallel executions and speed of the runs, I'd expect "no response" difficult to prove, especially when you'd need a broad stanza interception since you don't know what kind of response you shouldn't receive 😂

src/main/java/org/igniterealtime/smack/inttest/xep0352/CsiIntegrationTest.java
guusdk: Yeah, there's nothing that I can think of to assert the correct state of a client. The XEP seems to explicitly make this impossible.

guusdk:
Rebased.

PR #24, opened 2024-05-16
Release 1.2 of the project

guusdk:
Includes some tweaks that should go in prior to the release.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -83,7 +83,7 @@ jobs:
           java \
           -Dsinttest.service="example.org" \
           -Dsinttest.securityMode=disabled \
-          -Dsinttest.replyTimeout=60000 \
+          -Dsinttest.replyTimeout=5000 \
           -Dsinttest.adminAccountUsername=admin \
           -Dsinttest.adminAccountPassword=admin \
           -Dsinttest.enabledConnections=tcp \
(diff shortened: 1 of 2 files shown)


guusdk:
Note that this release declares a dependency on an unstable version of Smack: 4.5.0-alpha4-SNAPSHOT

As a result, the behavior of this version might change over time (if Smack releases newer variants of the same SNAPSHOT).

We're pressing on with this release anyway, as we'd like to benefit from changes that it brings. That said, this version should probably not be used by end-users, and should be regarded as an interim-release only.

.github/workflows/build.yml
```diff
@@ -83,13 +83,13 @@ jobs:
...
           -Dsinttest.securityMode=disabled \
-          -Dsinttest.replyTimeout=60000 \
+          -Dsinttest.replyTimeout=5000 \
           -Dsinttest.adminAccountUsername=admin \
           -Dsinttest.adminAccountPassword=admin \
           -Dsinttest.enabledConnections=tcp \
           -Dsinttest.dnsResolver=javax \
           -Dsinttest.disabledTests="EntityCapsTest,SoftwareInfoIntegrationTest,XmppConnectionIntegrationTest,StreamManagementTest,WaitForClosingStreamElementTest,IoTControlIntegrationTest,ModularXmppClientToServerConnectionLowLevelIntegrationTest" \
-          -Dsinttest.testRunResultProcessors=org.igniterealtime.smack.inttest.util.JUnitXmlTestRunResultProcessor \
+          -Dsinttest.testRunResultProcessors=org.igniterealtime.smack.inttest.SmackIntegrationTestFramework\$JulTestRunResultProcessor,org.igniterealtime.smack.inttest.util.JUnitXmlTestRunResultProcessor \
```
Fishbowler, line 92: This is right?

Fishbowler (approved):
Assuming that new config value is right, LGTM

.github/workflows/build.yml
guusdk: Yes, this is right.

Prior to having the entire line, the default test run result processor from Smack was used, generating output on std-out. When I added the JUnit XML-based runner, Smack's default processor was no longer used. With this change, I'm adding Smack's processor back, so that they now both run. Smack's default one is the one that generates the output that we're mentioning in https://xmpp-interop-testing.github.io/documentation/diagnose-test-failures - so it's good to have that.

PR #26, opened 2024-05-23
New tests for section "8.1 Moderator Use Cases: Modifying the Room Subject" of XEP-0045: "Multi-User Chat"

guusdk:
Note that these tests will fail when using a version of Smack that does not fix a bug in its subject change implementation. See https://github.com/igniterealtime/Smack/pull/596

### DIFF
--- a/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatModeratorSubjectModIntegrationTest.java
+++ b/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatModeratorSubjectModIntegrationTest.java
@@ -109,9 +109,9 @@ public class MultiUserChatModeratorSubjectModIntegrationTest extends AbstractMul
 
         createMuc(mucAsSeenByOne, nicknameOne);
         try {
-            final FormField field = mucAsSeenByOne.getConfigurationForm().getField("muc#roomconfig_changesubject");
-            if (field != null && ((BooleanFormField) field).getValueAsBoolean()) {
-                throw new TestNotPossibleException("Room is configured to allow subject changes from participants.");
+            final MucConfigFormManager configFormManager = mucAsSeenByOne.getConfigFormManager();
+            if (configFormManager.occupantsAreAllowedToChangeSubject()) {
+                configFormManager.disallowOccupantsToChangeSubject().submitConfigurationForm();
             }
             mucAsSeenByTwo.join(nicknameTwo);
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/jivesoftware/smackx/muc/MultiUserChatModeratorSubjectModIntegrationTest.java
```diff
@@ -0,0 +1,312 @@
...
+        }
+    }
+
+    /**
+     * Asserts that a non-moderator is <em>not</em> able to change the room subject.
+     */
+    @SmackIntegrationTest(section = "8.1", quote =
+        "If someone without appropriate privileges attempts to change the room subject, the service MUST return a " +
+        "message of type \"error\" specifying a <forbidden/> error condition")
+    public void mucTestParticipantNotAllowedToChangeSubject() throws Exception
```
Flowdalic, line 101: FWIW, https://github.com/igniterealtime/Smack/commit/6ae8234d255bcf6e19a5faf50d70a2892f35b0b4 is how I added this test. It has slight changes wrt to ensuring the feature is provided by the MUC service.

Also this test is called here 'mucTestParticipant…' and 'mucTestVisitor…' in Smack. I have to read up in XEP-0045 which one is more appropriate/correct (I assume participant?).

Flowdalic: FWIW, the commit mentioned above required https://github.com/igniterealtime/Smack/commit/a7083872101454f6dde10d6d69e546be3115ad9b. Ideally, most MUC configuration options would be included in `MucConfigFormManager` which can be seen as a high-level API for the data form-based MUC configuration.

Fishbowler: @Flowdalic The text on the commit at https://github.com/XMPP-Interop-Testing/smack-sint-server-extensions/pull/26/commits/01e032a6e03a78494771b444694141c5dc99ab8c gives Guus' rationale on one name over the other (that a Visitor role isn't required to be implemented by the server)

@guusdk If this is now included in Smack, should we remove it from this PR?

Fishbowler (commented):
I noticed that this doesn't include a test for the format stuff at the beginning of §8.1:

> The subject is changed by sending a message of type "groupchat" to the <room@service>, where the <message/> MUST contain a <subject/> element that specifies the new subject but MUST NOT contain a <body/> element (or a <thread/> element). In accordance with the core definition of XMPP, other child elements are allowed (although the entity that receives them might ignore them).
> 
> Note: A message with a <subject/> and a <body/> or a <subject/> and a <thread/> is a legitimate message, but it SHALL NOT be interpreted as a subject change.

Was this a choice, or was this missed?

src/main/java/org/jivesoftware/smackx/muc/MultiUserChatModeratorSubjectModIntegrationTest.java
guusdk: Given that Smack doesn't include all the tests from this set, I think it's preferable to have a 'complete set' in our code (as opposed to avoid duplication). When someone reviews our code for coverage, this particular scenario should not be missing.

I will refactor our tests to make use of the new API that was introduced in Smack.

guusdk:
> Was this a choice, or was this missed?

I'm not immediately sure how to write a reliable and maintainable test for this. Smack's API hides the details of the composition of the stanza. Should we compose a 'manual' message stanza? How do we check if the desired result unfolds itself?

guusdk:
Rebased

PR #29, opened 2024-06-14
Fix CSI test

guusdk:
These tests should not leave a connection in an 'inactive' state, as that will introduce weird behavior in tests executed later.

### DIFF
--- a/src/main/java/org/igniterealtime/smack/inttest/xep0352/CsiIntegrationTest.java
+++ b/src/main/java/org/igniterealtime/smack/inttest/xep0352/CsiIntegrationTest.java
@@ -23,7 +23,6 @@ import org.igniterealtime.smack.inttest.annotations.SpecificationReference;
 import org.igniterealtime.smack.inttest.util.IntegrationTestRosterUtil;
 import org.jivesoftware.smack.SmackException;
 import org.jivesoftware.smack.XMPPException;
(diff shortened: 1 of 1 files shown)


src/main/java/org/igniterealtime/smack/inttest/xep0352/CsiIntegrationTest.java
```diff
@@ -48,24 +46,26 @@ public CsiIntegrationTest(final SmackIntegrationTestEnvironment environment) thr
...
-     * @throws SmackException.NotConnectedException
-     * @throws InterruptedException
-     */
     @SmackIntegrationTest(section = "4.2", quote = "[...] when the client is active again it sends an <active/> element.")
-    public void markInactiveTest() throws SmackException.NotConnectedException, InterruptedException
+    public void markActiveTest() throws SmackException.NotConnectedException, InterruptedException
     {
-        ClientStateIndicationManager.inactive(conOne);
-        ClientStateIndicationManager.active(conOne);
-        // TODO do we need to reset the CSI-state of conOne to not affect other tests?
```
Fishbowler, line 68: 🤣

guusdk: Two days of my life.

PR #30, opened 2024-06-18
Add tests for XEP-0215 External Service Discovery

guusdk:
Note that for Openfire, none of these tests can run (they'll be automatically skipped) unless the [External Service Discovery plugin](https://www.igniterealtime.org/projects/openfire/plugin-archive.jsp?plugin=externalservicediscovery) is loaded.

At the time of writing, that plugin contains on bug that is correctly identified by the tests that I'm adding in this PR. The ticket for that bug is here: https://github.com/igniterealtime/openfire-externalservicediscovery-plugin/issues/16

### DIFF
--- a/src/main/java/org/igniterealtime/smack/inttest/xep0215/ExternalServiceDiscoveryIntegrationTest.java
+++ b/src/main/java/org/igniterealtime/smack/inttest/xep0215/ExternalServiceDiscoveryIntegrationTest.java
@@ -1,0 +1,184 @@
+/*
+ * Copyright 2024 Guus der Kinderen
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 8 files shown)


guusdk:
@Flowdalic there might be an opportunity to copy much of the stanza parsing from this project to Smack, to add XEP-0215 there. One thing that you probably do not want ported over is the string-representation of the 'expires' attribute. I'm exposing that as a raw string here (rather than as a Date object) to be able to assert that the value is in the XEP-defined format.

src/main/java/org/igniterealtime/smack/inttest/xep0215/ExternalServiceDiscoveryIntegrationTest.java
```diff
@@ -0,0 +1,184 @@
...
+        final DiscoverExternalServices request = new DiscoverExternalServices();
+        request.setTo(service);
+
+        final DiscoverExternalServices response = conOne.sendIqRequestAndWaitForResponse(request);
+        final Collection<String> discoveredExpiryDates = response.getServices().stream().map(DiscoverExternalServices.Service::getExpires).filter(Objects::nonNull).collect(Collectors.toSet());
+        if (discoveredExpiryDates.isEmpty()) {
+            throw new TestNotPossibleException("The server under test does not provide any data that has en 'expires' value.");
+        }
+
+        final String[] zuluEndings = new String[] { "z", "Z", "+00:00", "-00:00"};
```
Fishbowler, line 99: I wonder if there's some utlity in extracting this timestamp stuff out into a helper at some point. This isn't the first PR I've seen it in.

Fishbowler (approved):

PR #34, opened 2024-06-28
Release 1.3

guusdk:


guusdk:
Build failure caused by a bug in Openfire for which a fix is underway. I'll ignore it here.

PR #35, opened 2024-06-28
CI: Add Java 21 to list JDKs used for CI

guusdk:
Java 21 is a LTS release. Let's use that in our CI too.


Fishbowler (approved):
LGTM, test failures are correct VCard failures

PR #36, opened 2024-06-30
Enforce no presence subscription before testing presence pubsub model

Fishbowler:
This test is failing in Circle CI but not locally. I believe this to be a test ordering difference, and that a test that runs before this is subscribing conTwo to conOne, breaking the test.

I looked at tearing down the subscription in SubscriptionIntegrationTest, but this might not be where it's occuring, and this feels more explicit.

### DIFF
--- a/src/main/java/org/igniterealtime/smack/inttest/xep0115/EntityCapabilitiesOptimizationIntegrationTest.java
+++ b/src/main/java/org/igniterealtime/smack/inttest/xep0115/EntityCapabilitiesOptimizationIntegrationTest.java
@@ -90,6 +90,7 @@ public class EntityCapabilitiesOptimizationIntegrationTest extends AbstractSmack
         } finally {
             // Tear down test fixture.
             rosterOne.removePresenceEventListener(presenceEventListener);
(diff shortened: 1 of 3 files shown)


guusdk:
I think this is a good improvement, but we must also find the test that doesn't properly clean up after itself (necessitating this change). If we leave the offender unchecked, it'll only re-introduce a similar issue with another test in the future.

guusdk:
Maybe a good start is to ensure that all tests where "ensure users are in each-others rosters" are followed up with a 'test fixture tear-down step' where that is being undone.

guusdk:
Related improvements for Smack: https://github.com/igniterealtime/Smack/pull/605

guusdk:
I've added similar changes to this PR. @Fishbowler can you have a quick review? If you're happy, I'm happy to merge this as well.

guusdk (approved):

Fishbowler:
I had a similar change locally, but decided it might be overkill, and considered that the setup of the state of a test was the purview of the test that's running. No objections though.

guusdk:
I believe that it is sensible for a test to leave the system in the state that it was originally in, pre-test. That reduces the risk of tests affecting each-other.

PR #38, opened 2024-07-03
CI: Use Openfire 4.8.2 instead of Openfire 4.8.1

guusdk:
In Openfire 4.8.2, improvements have been applied that make more of our tests go green.


guusdk:
Test fails as Openfire 4.8.2 has not been released yet.

guusdk:
All green now.

PR #41, opened 2024-08-15
Add tests for XEP-0045 Multi-User Chat (section 10)

guusdk:
This contains all tests for section 10 "Owner Use Cases" of XEP-0045.

Some of the tests included here are completely or partially commented out, as they depend on fixes in the underlying library, Smack, such as:
- [SMACK-947](https://igniterealtime.atlassian.net/browse/SMACK-947) / https://github.com/igniterealtime/Smack/pull/614
- [SMACK-949](https://igniterealtime.atlassian.net/browse/SMACK-949)
- [SMACK-950](https://igniterealtime.atlassian.net/browse/SMACK-950) / https://github.com/igniterealtime/Smack/pull/616

### DIFF
--- a/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatOwnerAdminListIntegrationTest.java
+++ b/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatOwnerAdminListIntegrationTest.java
@@ -1,0 +1,591 @@
+/**
+ * Copyright 2024 Guus der Kinderen
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jivesoftware.smackx.muc;
+
+import org.igniterealtime.smack.inttest.SmackIntegrationTestEnvironment;
+import org.igniterealtime.smack.inttest.TestNotPossibleException;
+import org.igniterealtime.smack.inttest.annotations.SmackIntegrationTest;
+import org.igniterealtime.smack.inttest.annotations.SpecificationReference;
+import org.igniterealtime.smack.inttest.util.SimpleResultSyncPoint;
+import org.jivesoftware.smack.SmackException;
+import org.jivesoftware.smack.XMPPException;
+import org.jivesoftware.smack.packet.IQ;
+import org.jivesoftware.smack.packet.StanzaError;
+import org.jivesoftware.smackx.muc.packet.MUCAdmin;
+import org.jivesoftware.smackx.muc.packet.MUCItem;
+import org.jxmpp.jid.EntityBareJid;
+import org.jxmpp.jid.EntityFullJid;
+import org.jxmpp.jid.impl.JidCreate;
+import org.jxmpp.jid.parts.Resourcepart;
+import org.jxmpp.stringprep.XmppStringprepException;
+
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+/**
+ * Tests for section "10.8 Owner Use Cases: Modifying the Admin List" of XEP-0045: "Multi-User Chat"
+ *
+ * @see <a href="https://xmpp.org/extensions/xep-0045.html#modifyadmin">XEP-0045 Section 10.8</a>
+ */
+@SpecificationReference(document = "XEP-0045", version = "1.34.6")
+public class MultiUserChatOwnerAdminListIntegrationTest extends AbstractMultiUserChatIntegrationTest
+{
+    public MultiUserChatOwnerAdminListIntegrationTest(SmackIntegrationTestEnvironment environment)
+        throws SmackException.NoResponseException, XMPPException.XMPPErrorException,
+        SmackException.NotConnectedException, InterruptedException, TestNotPossibleException, XmppStringprepException, MultiUserChatException.MucAlreadyJoinedException, MultiUserChatException.MissingMucCreationAcknowledgeException, MultiUserChatException.NotAMucServiceException
+    {
+        super(environment);
+
+        // The specification reads (in section 5.2) "Support for the admin [...] is RECOMMENDED."
+        // which suggests that this is optional functionality. The specification does not explicitly say how to test for
+        // support. This implementation will use any XMPP error in response to a change request as an indication that
+        // the feature is not supported by the server under test.
+        final EntityBareJid mucAddress = getRandomRoom("smack-inttest-owner-admin-revoke-support");
+        final MultiUserChat muc = mucManagerOne.getMultiUserChat(mucAddress);
+        createMuc(muc, Resourcepart.from("owner-" + randomString));
+        try {
+            muc.grantAdmin(conTwo.getUser().asBareJid());
+            muc.revokeAdmin(conTwo.getUser().asEntityBareJid());
+        } catch (XMPPException.XMPPErrorException e) {
+            if (StanzaError.Condition.not_authorized.equals(e.getStanzaError().getCondition())) {
+                throw new TestNotPossibleException("Service does not support granting and/or revokation of admin status functionality.");
+            } else {
+                throw e;
+            }
+        } finally {
+            tryDestroy(muc);
+        }
+    }
+
+    /**
+     * Asserts that an admin list can be obtained.
+     */
+    @SmackIntegrationTest(section = "10.8", quote = "the owner [...] requests the admin list by querying the room for all users with an affiliation of 'admin'.")
+    public void testOwnerRequestsAdminList() throws Exception
+    {
+        // Setup test fixture.
+        final EntityBareJid mucAddress = getRandomRoom("smack-inttest-owner-requests-adminlist");
+        final MultiUserChat mucAsSeenByOwner = mucManagerOne.getMultiUserChat(mucAddress);
+
+        final Resourcepart nicknameOwner = Resourcepart.from("owner-" + randomString);
+
+        createMuc(mucAsSeenByOwner, nicknameOwner);
+        try {
+            // Execute system under test.
+            final MUCAdmin iq = new MUCAdmin();
+            iq.setTo(mucAddress);
+            iq.setType(IQ.Type.get);
+            iq.addItem(new MUCItem(MUCAffiliation.admin));
+            conOne.sendIqRequestAndWaitForResponse(iq);
+
+            // Verify result.
+        } catch (XMPPException.XMPPErrorException e) {
+            fail("Expected owner '" + conOne.getUser() + "' to be able to receive the admin list from '" + mucAddress + "' (but the server returned an error).", e);
+        } finally {
+            // Tear down test fixture.
+            tryDestroy(mucAsSeenByOwner);
+        }
+    }
+
+    /**
+     * Asserts that a user (not in the room, that is semi-anonymous) cannot request the admin list.
+     *
+     * This test uses a semi-anonymous room, as a XEP update is in the works that allows these requests for non-anonymous rooms.
+     */
+    @SmackIntegrationTest(section = "10.8", quote = "If the <user@host> of the 'from' address does not match the bare JID of a room owner, the service MUST return a <forbidden/> error to the sender.")
+    public void testUserRequestsAdminList() throws Exception
+    {
+        // Setup test fixture.
+        final EntityBareJid mucAddress = getRandomRoom("smack-inttest-owner-user-requests-adminlist");
+        final MultiUserChat mucAsSeenByOwner = mucManagerOne.getMultiUserChat(mucAddress);
+
+        final Resourcepart nicknameOwner = Resourcepart.from("owner-" + randomString);
+
+        createSemiAnonymousMuc(mucAsSeenByOwner, nicknameOwner);
+        try {
+            // Execute system under test.
+            final MUCAdmin iq = new MUCAdmin();
+            iq.setTo(mucAddress);
+            iq.setType(IQ.Type.get);
+            iq.addItem(new MUCItem(MUCAffiliation.admin));
+
+            // Verify result.
+            final XMPPException.XMPPErrorException e = assertThrows(XMPPException.XMPPErrorException.class, () -> conTwo.sendIqRequestAndWaitForResponse(iq),
+                "Expected user '" + conTwo.getUser() + "' (that is not in the room) to receive an error when requesting the admin list from '" + mucAddress + "' (but the server did not return an error).");
+            assertEquals(StanzaError.Condition.forbidden, e.getStanzaError().getCondition(), "Unexpected error condition in the (expected) error that was returned to '" + conTwo.getUser() + "' (that is not in the room) after it requested the admin list of room '" + mucAddress + "'.");
+        } finally {
+            // Tear down test fixture.
+            tryDestroy(mucAsSeenByOwner);
+        }
+    }
+
+    /**
+     * Asserts that a participant (that is in a semi-anonymous room) cannot request the admin list.
+     *
+     * This test uses a semi-anonymous room, as a XEP update is in the works that allows these requests for non-anonymous rooms.
+     */
+    @SmackIntegrationTest(section = "10.8", quote = "If the <user@host> of the 'from' address does not match the bare JID of a room owner, the service MUST return a <forbidden/> error to the sender.")
+    public void testParticipantRequestsAdminList() throws Exception
+    {
+        // Setup test fixture.
+        final EntityBareJid mucAddress = getRandomRoom("smack-inttest-owner-user-requests-adminlist");
+        final MultiUserChat mucAsSeenByOwner = mucManagerOne.getMultiUserChat(mucAddress);
+        final MultiUserChat mucAsSeenByParticipant = mucManagerTwo.getMultiUserChat(mucAddress);
+
+        final Resourcepart nicknameOwner = Resourcepart.from("owner-" + randomString);
+        final Resourcepart nicknameParticipant = Resourcepart.from("participant-" + randomString);
+
+        createSemiAnonymousMuc(mucAsSeenByOwner, nicknameOwner);
+        try {
+            mucAsSeenByParticipant.join(nicknameParticipant);
+
+            // Execute system under test.
+            final MUCAdmin iq = new MUCAdmin();
+            iq.setTo(mucAddress);
+            iq.setType(IQ.Type.get);
+            iq.addItem(new MUCItem(MUCAffiliation.admin));
+
+            // Verify result.
+            final XMPPException.XMPPErrorException e = assertThrows(XMPPException.XMPPErrorException.class, () -> conTwo.sendIqRequestAndWaitForResponse(iq),
+                "Expected participant '" + conTwo.getUser() + "' to receive an error when requesting the admin list from '" + mucAddress + "' (but the server did not return an error).");
+            assertEquals(StanzaError.Condition.forbidden, e.getStanzaError().getCondition(), "Unexpected error condition in the (expected) error that was returned to '" + conTwo.getUser() + "' (that is a participant in the room) after it requested the admin list of room '" + mucAddress + "'.");
+        } finally {
+            // Tear down test fixture.
+            tryDestroy(mucAsSeenByOwner);
+        }
+    }
+
+    /**
+     * Asserts that an admin list item has 'affiliation' and 'jid' attributes.
+     */
+    @SmackIntegrationTest(section = "10.8", quote = "each item MUST include the 'affiliation' and 'jid' attributes")
+    public void testAdminListItemCheck() throws Exception
+    {
+        // Setup test fixture.
+        final EntityBareJid mucAddress = getRandomRoom("smack-inttest-owner-admin-list-itemcheck");
+        final MultiUserChat mucAsSeenByOwner = mucManagerOne.getMultiUserChat(mucAddress);
+        final MultiUserChat mucAsSeenByTarget = mucManagerTwo.getMultiUserChat(mucAddress);
+
+        final Resourcepart nicknameOwner = Resourcepart.from("owner-" + randomString);
+        final Resourcepart nicknameTarget = Resourcepart.from("target-" + randomString);
+
+        createMuc(mucAsSeenByOwner, nicknameOwner);
+        try {
+            mucAsSeenByOwner.grantAdmin(conTwo.getUser().asBareJid());
+            mucAsSeenByTarget.join(nicknameTarget);
+
+            // Execute system under test.
+            final MUCAdmin iq = new MUCAdmin();
+            iq.setTo(mucAddress);
+            iq.setType(IQ.Type.get);
+            iq.addItem(new MUCItem(MUCAffiliation.admin));
+            final MUCAdmin response = conOne.sendIqRequestAndWaitForResponse(iq);
+
+            // Verify result.
+            assertFalse(response.getItems().stream().anyMatch(i -> i.getAffiliation() == null), "The admin list for '" + mucAddress + "' contains an item that does not have an 'affiliation' attribute (but all items must have one).");
+            assertFalse(response.getItems().stream().anyMatch(i -> i.getJid() == null), "The admin list for '" + mucAddress + "' contains an item that does not have a 'jid' attribute (but all items must have one).");
+            assertTrue(response.getItems().stream().anyMatch(i -> i.getJid().equals(conTwo.getUser().asBareJid())), "Expected the admin list requested by '" + conOne.getUser() + "' from '" + mucAddress + "' to include '" + conTwo.getUser().asBareJid() + "' (but it did not).");
+        } finally {
+            // Tear down test fixture.
+            tryDestroy(mucAsSeenByOwner);
+        }
+    }
+
+    /**
+     * Asserts that an admin list modification can contain more than one item.
+     */
+    @SmackIntegrationTest(section = "10.8", quote = "The owner can then modify the admin list if desired. In order to do so, the owner MUST send the changed items [...] back to the service [...] the service MUST modify admin list and then inform the owner of success")
+    public void testAdminListMultipleItems() throws Exception
+    {
+        // Setup test fixture.
+        final EntityBareJid mucAddress = getRandomRoom("smack-inttest-owner-adminlist-multiple");
+        final MultiUserChat mucAsSeenByOwner = mucManagerOne.getMultiUserChat(mucAddress);
+        final Resourcepart nicknameOwner = Resourcepart.from("owner-" + randomString);
+
+        createMuc(mucAsSeenByOwner, nicknameOwner);
+        try {
+            // Execute system under test.
+            final MUCAdmin iq = new MUCAdmin();
+            iq.setTo(mucAddress);
+            iq.setType(IQ.Type.set);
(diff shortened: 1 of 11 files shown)


Fishbowler:
What's the solution here? Comment more out to get tests passing in the short term?

guusdk:
> What's the solution here? Comment more out to get tests passing in the short term?

Commenting out tests should only be done if they're not (yet) compatible with either the specification or with the implementation Smack. Openfire not being compliant with the specification should not be a reason to comment out the test (although we _may_ consider not running that particular test for Openfire by modifying our CI setup).

Ideally, we verify that the tests are correct (even if Openfire is not). In parallel, I'm trying to get fixes in Openfire (but those are out of scope of this project).

guusdk:
I've rebased this PR to the latest head.

Fishbowler:
Are the remaining failures all Openfire bugs?

guusdk:
I think so. I've got a rewrite of Openfire in the works on which all of the tests seem to pass.

guusdk:
The test failures in CI are fixed by https://github.com/igniterealtime/Openfire/pull/2505 and https://github.com/igniterealtime/Openfire/pull/2510 which will go in a future release of Openfire (likely 4.10.0)

guusdk:
Rebased.

guusdk:
Rebased, and enabled tests that were commented out pending changes in Smack (that have now landed).

src/main/java/org/jivesoftware/smackx/muc/MultiUserChatOwnerAdminListIntegrationTest.java
```diff
@@ -0,0 +1,591 @@
...
+            // Execute system under test.
+            final MUCAdmin iq = new MUCAdmin();
+            iq.setTo(mucAddress);
+            iq.setType(IQ.Type.get);
+            iq.addItem(new MUCItem(MUCAffiliation.admin));
+            final MUCAdmin response = conOne.sendIqRequestAndWaitForResponse(iq);
+
+            // Verify result.
+            assertFalse(response.getItems().stream().anyMatch(i -> i.getAffiliation() == null), "The admin list for '" + mucAddress + "' contains an item that does not have an 'affiliation' attribute (but all items must have one).");
+            assertFalse(response.getItems().stream().anyMatch(i -> i.getJid() == null), "The admin list for '" + mucAddress + "' contains an item that does not have a 'jid' attribute (but all items must have one).");
```
Fishbowler, line 200: Do we have guarantees from Smack that these will be null, rather than empty?

src/main/java/org/jivesoftware/smackx/muc/MultiUserChatOwnerAdminListIntegrationTest.java
```diff
@@ -0,0 +1,591 @@
...
+            // Tear down test fixture.
+            tryDestroy(mucAsSeenByOwner);
+        }
+    }
+
+    /**
+     * Asserts that an admin list modification is a delta: it shouldn't affect entries already on the admin list
+     * that are not included in the delta.
+     */
+    @SmackIntegrationTest(section = "10.8", quote = "The owner can then modify the admin list if desired. In order to do so, the owner MUST send the changed items (i.e., only the \"delta\")")
```
Fishbowler, line 289: Should this test be checking that "something not in the delta was unmodified", rather than "items that were in the delta were modified"?

src/main/java/org/jivesoftware/smackx/muc/MultiUserChatOwnerAdminListIntegrationTest.java
```diff
@@ -0,0 +1,591 @@
...
+        final EntityFullJid target1MucAddress = JidCreate.entityFullFrom(mucAddress, nicknameTarget1);
+        final EntityFullJid target2MucAddress = JidCreate.entityFullFrom(mucAddress, nicknameTarget2);
+
+        createMuc(mucAsSeenByOwner, nicknameOwner);
+        try {
+            try {
+                mucAsSeenByOwner.grantAdmin(conTwo.getUser().asBareJid());
+            } catch (XMPPException.XMPPErrorException e) {
+                throw new TestNotPossibleException("Unable to grant '" + conTwo.getUser().asBareJid() + "' admin status in room '" + mucAddress + "'.");
+            }
```
Fishbowler, line 493: Shouldn't this be impossible, given the stuff in the setup?

Fishbowler (commented):
A few comments after reviewing the Admin List tests.

src/main/java/org/jivesoftware/smackx/muc/MultiUserChatOwnerConfigRoomIntegrationTest.java
```diff
@@ -0,0 +1,1018 @@
...
+            for (final FormField field : configForm.getDataForm().getFields()) {
+                // Set any non-default value for list-single fields
+                if (field instanceof ListSingleFormField) {
+                    final List<FormField.Option> options = ((ListSingleFormField) field).getOptions();
+                    final FormField.Value defaultValue = ((ListSingleFormField) field).getRawValue();
+                    if (!options.isEmpty()) {
+                        for (final FormField.Option option : options) {
+                            if (!option.getValue().equals(defaultValue)) {
+                                optionValues.put(field.getFieldName(), option.getValue());
+                                answerForm.setAnswer(field.getFieldName(), option.getValue().getValue());
```
Fishbowler, line 241: Is `.getValue().getValue()` correct?

Fishbowler (commented):
Comments after reviewing the Owner Config tests.

src/main/java/org/jivesoftware/smackx/muc/MultiUserChatOwnerCreateRoomIntegrationTest.java
```diff
@@ -0,0 +1,547 @@
...
+        requestConfigurationFormStanza.setType(IQ.Type.get);
+
+        try (final StanzaCollector collector = conOne.createStanzaCollectorAndSend(responseFilter, createMucStanza)) {
+            collector.nextResult();
+
+            // Execute system under test.
+            final IQ response = conOne.sendIqRequestAndWaitForResponse(requestConfigurationFormStanza);
+
+            // Verify result.
+            assertTrue(response instanceof MUCOwner || response.hasExtension(MUCOwner.ELEMENT, MUCOwner.NAMESPACE), "Expected the response with stanza ID '" + response.getStanzaId() + "' from '" + mucAddress + "' to the request for a configuration form from '" + conOne.getUser() + "' (MUC address '" + ownerMucAddress + "') to be a stanza that includes a child element named '" + MUCOwner.ELEMENT + "' qualified by the namespace '" + MUCOwner.NAMESPACE + "' (but it does not).");
```
Fishbowler, line 198: Is the first part of the or necessary if we've already got the second part?

Fishbowler (commented):
Comment on Create Room tests

src/main/java/org/jivesoftware/smackx/muc/MultiUserChatOwnerDestroyRoomIntegrationTest.java
```diff
@@ -0,0 +1,434 @@
...
+            } catch (XMPPException.XMPPErrorException e) { // TODO remove this catch after SMACK-949 gets fixed.
+                // Room was likely already destroyed.
+            }
+        }
+    }
+
+    /**
+     * Verifies that a room destruction request cannot be issued by a non-owner.
+     */
+    @SmackIntegrationTest(section = "10.9", quote = "If the <user@host> of the 'from' address received on a destroy request does not match the bare JID of a room owner, the service MUST return a <forbidden/> error to the sender")
```
Fishbowler, line 398: There's an argument that we could extend this to particpant and admin too.

src/main/java/org/jivesoftware/smackx/muc/MultiUserChatOwnerGrantAdminIntegrationTest.java
```diff
@@ -0,0 +1,495 @@
...
+            tryDestroy(muc);
+        }
+    }
+
+    /**
+     * Verifies that an owner can grant admin status to a user that is currently not in the room.
+     */
+    @SmackIntegrationTest(section = "10.6", quote = "An owner can grant admin status to a member or an unaffiliated user; this is done by changing the user's affiliation to \"admin\"")
+    public void testGrantAdmin() throws Exception
+    {
```
Fishbowler, line 72: Does this need an explicit test for members?

src/main/java/org/jivesoftware/smackx/muc/MultiUserChatOwnerGrantAdminIntegrationTest.java
```diff
@@ -0,0 +1,495 @@
...
+            } catch (XMPPException.XMPPErrorException e) {
+                fail("Expected owner '" + conOne.getUser() + "' to be able to grant admin status to '" + conTwo.getUser().asBareJid() + "' (that is currently joined as '" + nicknameTarget+ "') in '" + mucAddress + "' (but the server returned an error).", e);
+            }
+        } finally {
+            // Tear down test fixture.
+            tryDestroy(mucAsSeenByOwner);
+        }
+    }
+
+    // TODO enable these tests after https://github.com/xsf/xeps/pull/1370 gets merged. Until then, the specification does not seem to restrict granting of admin status to owners.
```
Fishbowler, line 225: Clarification, to avoid confusion with "making an owner an admin" which was my initial interpretation (which, given it's affiliation based, would be downgrading, which is something we sort-of test elsewhere)
```suggestion
    // TODO enable these tests after https://github.com/xsf/xeps/pull/1370 gets merged. Until then, the specification does not seem to restrict granting of admin as something only owners can do.
```

Fishbowler (commented):
Comments on Grant Admin tests

src/main/java/org/jivesoftware/smackx/muc/MultiUserChatOwnerGrantOwnerIntegrationTest.java
```diff
@@ -0,0 +1,495 @@
...
+            tryDestroy(mucAsSeenByOwner);
+        }
+    }
+
+    /**
+     * Verifies that an owner can grant owner status to a user that is currently a participant in the room, while providing a 'reason'.
+     */
+    @SmackIntegrationTest(section = "10.3", quote = "an owner MAY grant owner status to another user; this is done by changing the user's affiliation to \"owner\" [...] The <reason/> element is OPTIONAL.")
+    public void testGrantOwnerOptionalReasonWhileInRoom() throws Exception
+    {
```
Fishbowler, line 182: We test reasons and we test presence, but don't ever test that the presence includes reasons. Is that worth a TODO?

Fishbowler (commented):
Comment on Grant Owner tests

src/main/java/org/jivesoftware/smackx/muc/MultiUserChatOwnerGrantOwnerIntegrationTest.java
Fishbowler: I'll bet the same applies to Grant Admin tests.

Fishbowler (commented):
No notes on Chat Owner tests :)

src/main/java/org/jivesoftware/smackx/muc/MultiUserChatOwnerOwnerListIntegrationTest.java
```diff
@@ -0,0 +1,617 @@
...
+            // Execute system under test.
+            final MUCAdmin iq = new MUCAdmin();
+            iq.setTo(mucAddress);
+            iq.setType(IQ.Type.get);
+            iq.addItem(new MUCItem(MUCAffiliation.owner));
+            final MUCAdmin response = conOne.sendIqRequestAndWaitForResponse(iq);
+
+            // Verify result.
+            assertFalse(response.getItems().stream().anyMatch(i -> i.getAffiliation() == null), "The owner list for '" + mucAddress + "' contains an item that does not have an 'affiliation' attribute (but all items must have one).");
+            assertFalse(response.getItems().stream().anyMatch(i -> i.getJid() == null), "The owner list for '" + mucAddress + "' contains an item that does not have a 'jid' attribute (but all items must have one).");
```
Fishbowler, line 197: As with the Admin List, does Smack guarantee that the JID will either be null or useful?

src/main/java/org/jivesoftware/smackx/muc/MultiUserChatOwnerOwnerListIntegrationTest.java
```diff
@@ -0,0 +1,617 @@
...
+
+        final Resourcepart nicknameOwner = Resourcepart.from("owner-" + randomString);
+
+        createMuc(mucAsSeenByOwner, nicknameOwner);
+        try {
+            try {
+                mucAsSeenByOwner.grantOwnership(List.of(conTwo.getUser().asBareJid(), conThree.getUser().asBareJid()));
+            } catch (XMPPException.XMPPErrorException e) {
+                throw new TestNotPossibleException("Unable to grant '" + conTwo.getUser().asBareJid() + "' and/or '" + conThree.getUser().asBareJid() + "' owner status in room '" + mucAddress + "'.");
+            }
```
Fishbowler, line 258: Is this redundant, given the checking done in the constructor?

src/main/java/org/jivesoftware/smackx/muc/MultiUserChatOwnerOwnerListIntegrationTest.java
```diff
@@ -0,0 +1,617 @@
...
+
+            try {
+                conOne.sendIqRequestAndWaitForResponse(iq);
+
+                // Verify result.
+            } catch (XMPPException.XMPPErrorException e) {
+                fail("Expected the service to inform '" + conOne.getUser() + "' of success after they modified the owner list of room '" + mucAddress + "' (but instead, an error was returned).", e);
+            }
+            assertTrue(mucAsSeenByOwner.getOwners().stream().noneMatch(i -> i.getAffiliation() == MUCAffiliation.owner && i.getJid().equals(conTwo.getUser().asBareJid())), "Expected the owner list for '" + mucAddress + "' to not contain '" + conTwo.getUser().asBareJid() + "' that was just removed from the owner list by '" + conOne.getUser() + "' (but does appear on the owner list).");
+            assertTrue(mucAsSeenByOwner.getOwners().stream().noneMatch(i -> i.getAffiliation() == MUCAffiliation.owner && i.getJid().equals(conThree.getUser().asBareJid())), "Expected the owner list for '" + mucAddress + "' to not contain '" + conThree.getUser().asBareJid() + "' that was just removed from the owner list by '" + conOne.getUser() + "' (but does appear on the owner list).");
```
Fishbowler, line 275: Are these checks overcomplicated? Shouldn't this just be the absence of these users from the list of owners? Can the Owners list ever return people with a non-owner affiliation?

Fishbowler (commented):
Comments on Owner List tests

src/main/java/org/jivesoftware/smackx/muc/MultiUserChatOwnerRevokeAdminIntegrationTest.java
```diff
@@ -0,0 +1,577 @@
...
+    public MultiUserChatOwnerRevokeAdminIntegrationTest(SmackIntegrationTestEnvironment environment)
+        throws SmackException.NoResponseException, XMPPException.XMPPErrorException,
+        SmackException.NotConnectedException, InterruptedException, TestNotPossibleException, XmppStringprepException, MultiUserChatException.MucAlreadyJoinedException, MultiUserChatException.MissingMucCreationAcknowledgeException, MultiUserChatException.NotAMucServiceException
+    {
+        super(environment);
+
+        // The specification reads (in section 5.2) "Support for the admin [...] is RECOMMENDED."
+        // which suggests that this is optional functionality. The specification does not explicitly say how to test for
+        // support. This implementation will use any XMPP error in response to a change request as an indication that
+        // the feature is not supported by the server under test.
```
Fishbowler, line 55: I love this.

src/main/java/org/jivesoftware/smackx/muc/MultiUserChatOwnerRevokeAdminIntegrationTest.java
```diff
@@ -0,0 +1,577 @@
...
+                fail("Expected owner '" + conOne.getUser() + "' to be able to revoke admin status from '" + conTwo.getUser().asBareJid() + "' (that is currently joined as '" + nicknameTarget+ "') in '" + mucAddress + "' (but the server returned an error).", e);
+            }
+        } finally {
+            // Tear down test fixture.
+            tryDestroy(mucAsSeenByOwner);
+        }
+    }
+
+    /**
+     * Verifies that an owner can revoke admin status from a user that is currently not in the room.
```
Fishbowler, line 168: Add disambiguation from the earlier test
```suggestion
     * Verifies that an owner can revoke admin status from a user that is currently not in the room, and give a reason.
```

src/main/java/org/jivesoftware/smackx/muc/MultiUserChatOwnerRevokeAdminIntegrationTest.java
```diff
@@ -0,0 +1,577 @@
...
+                fail("Expected owner '" + conOne.getUser() + "' to be able to revoke admin status from '" + conTwo.getUser().asBareJid() + "' in '" + mucAddress + "' (but the server returned an error).", e);
+            }
+        } finally {
+            // Tear down test fixture.
+            tryDestroy(mucAsSeenByOwner);
+        }
+    }
+
+    /**
+     * Verifies that an owner can revoke admin status from a user that is currently a participant in the room.
```
Fishbowler, line 206: Disambiguation
```suggestion
     * Verifies that an owner can revoke admin status from a user that is currently a participant in the room, and provide a reason.
```

src/main/java/org/jivesoftware/smackx/muc/MultiUserChatOwnerRevokeAdminIntegrationTest.java
```diff
@@ -0,0 +1,577 @@
...
+            grantRequest.setType(IQ.Type.set);
+            grantRequest.addItem(new MUCItem(MUCAffiliation.admin, conTwo.getUser().asBareJid()));
+
+            conOne.sendIqRequestAndWaitForResponse(grantRequest);
+
+            // Execute system under test.
+            final MUCAdmin revokeRequest = new MUCAdmin();
+            revokeRequest.setTo(mucAddress);
+            revokeRequest.setType(IQ.Type.set);
+            revokeRequest.addItem(new MUCItem(MUCAffiliation.owner, conTwo.getUser().asBareJid(), "Revoking Admin Status as part of an integration test."));
```
Fishbowler, line 246: As with some other tests, I was going to comment that we could add a TODO for testing that users are notified of the reason.
Except that XEP-0045 s10.7 says that the reason is optional, but not how to route it (unlike, for example, 10.9, where room destruction reasons must be relayed).

Fishbowler (commented):
Comments on Revoke Admin tests

Fishbowler (commented):
No comments on Revoke Owner tests :)

Fishbowler (approved):
I've left a lot of comments, but most are prompts or questions. Assuming none of them lead to changes, am definitely happy to merge as-is.

src/main/java/org/jivesoftware/smackx/muc/MultiUserChatOwnerAdminListIntegrationTest.java
```diff
@@ -0,0 +1,591 @@
...
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+/**
+ * Tests for section "10.8 Owner Use Cases: Modifying the Admin List" of XEP-0045: "Multi-User Chat"
+ *
+ * @see <a href="https://xmpp.org/extensions/xep-0045.html#modifyadmin">XEP-0045 Section 10.8</a>
+ */
+@SpecificationReference(document = "XEP-0045", version = "1.34.6")
```
Fishbowler, line 44: We should update this to `1.35.0` throughout though.

src/main/java/org/jivesoftware/smackx/muc/MultiUserChatOwnerAdminListIntegrationTest.java
guusdk: Yup. They are both a specific type, not a string. That means that they can't be 'empty' like a String can.

src/main/java/org/jivesoftware/smackx/muc/MultiUserChatOwnerAdminListIntegrationTest.java
guusdk: Yes! I got confused by a couple of competing requirements:
- the request _needs_ to have at least two changes (otherwise it would not be a 'change list' request, but rather an 'add' or 'remove' for an individual entity
- we only have three connections to play with (which is a limitation in the SINT framework)
- I do not know if we can expect servers to allow admins that are not local users to be added.

I have now changed this test to add a (non-local) admin in the test setup, that's then not changed by the test. An assertion is added that this admin remains to be an admin after the system-under-test is invoked. Finally, a TestNotPossible is thrown if the server doesn't allow this user to be an admin in the first place.

src/main/java/org/jivesoftware/smackx/muc/MultiUserChatOwnerAdminListIntegrationTest.java
guusdk: I believe that I've added this check ('see if we can make this entity an admin, if not, throw TestNotPossible') consistently through the code. I've done that because the XEP is not strict in the 'admin' affiliation to be required. Section 5.2 reads:
> Support for the owner affiliation is REQUIRED. Support for the admin, member, and outcast affiliations is RECOMMENDED. (The "None" affiliation is the absence of an affiliation.)

This is my take at coping with a server that doesn't support the 'admin' affiliation (although I never tested it, and fully expect a great deal of weird testing behavior if we actually run against a server that does indeed not support this affiliation. I'm open for suggestions...

src/main/java/org/jivesoftware/smackx/muc/MultiUserChatOwnerConfigRoomIntegrationTest.java
```diff
@@ -0,0 +1,1018 @@
...
+        final Resourcepart nicknameOwner = Resourcepart.from("owner-" + randomString);
+
+        final Map<String, FormField.Value> optionValues = new HashMap<>();
+        try {
+            mucAsSeenByOwner.create(nicknameOwner);
+            final Form configForm = mucAsSeenByOwner.getConfigurationForm();
+            final FillableForm answerForm = configForm.getFillableForm();
+
+            for (final FormField field : configForm.getDataForm().getFields()) {
+                // Set any non-default value for list-single fields
```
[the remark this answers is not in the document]
guusdk: I don't think it matters if it's the last or the first non-default value - any non default value will do. I will add a break after the first value is set to the form - no need to iterate further if a non-default value was found.

src/main/java/org/jivesoftware/smackx/muc/MultiUserChatOwnerConfigRoomIntegrationTest.java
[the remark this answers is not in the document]
guusdk: Yes. The first `getValue` gets us a 'Value' instance. The second `getValue` gives us the textual representation of that instance, which is what we need to set the answer.

src/main/java/org/jivesoftware/smackx/muc/MultiUserChatOwnerCreateRoomIntegrationTest.java
guusdk: I'm not sure if Smack will _always_ be able to serialize the response as a MUCOwner. If it doesn't, then the second condition still verifies that the assertion is met. If it does, however, then I've observed the second condition to be `false`.

src/main/java/org/jivesoftware/smackx/muc/MultiUserChatOwnerDestroyRoomIntegrationTest.java
guusdk: Agreed. I've added tests for all types of affiliation.

src/main/java/org/jivesoftware/smackx/muc/MultiUserChatOwnerGrantAdminIntegrationTest.java
guusdk: Agreed. I've added that, as well as for the variant in which the user has joined the room.

src/main/java/org/jivesoftware/smackx/muc/MultiUserChatOwnerGrantOwnerIntegrationTest.java
guusdk: The presence that is broadcast to the target and possibly the other occupants, you mean? I do not think that the XEP mandates that this reason MUST be in that presence, strictly speaking. (eg: perhaps it's only used for auditing).

I'd _expect_ that reason to be broadcast, but I don't think it's a requirement in the XEP.

src/main/java/org/jivesoftware/smackx/muc/MultiUserChatOwnerOwnerListIntegrationTest.java
guusdk: Yup - these are not strings, thus cannot be 'empty'.

src/main/java/org/jivesoftware/smackx/muc/MultiUserChatOwnerOwnerListIntegrationTest.java
guusdk: Strictly speaking, I think you're right. This construct is copied all over, that I'd prefer to leave this in as a bit of (redundant) safety though.

src/main/java/org/jivesoftware/smackx/muc/MultiUserChatOwnerOwnerListIntegrationTest.java
guusdk: I don't think they should, but does the explicit check hurt? I'm not sure if requesting the owner list MAY include other affiliations (feels wrong, but ... unsure).

src/main/java/org/jivesoftware/smackx/muc/MultiUserChatOwnerAdminListIntegrationTest.java
guusdk: Agreed, but there's another change to the XEP coming. I'd prefer to do all that in one go.

guusdk:
@Fishbowler thanks for the review! I've pushed a new commit that applies various changes based on your feedback. I have left conversations in this PR unresolved if I'd like to get your feedback on my response. Can you either comment on my comment (on your comment), and/or resolve these please? If everything is resolved, lets hit the merge button!

guusdk:
Lots of test failures in this PR, but when I execute the locally against [a version of Openfire that has MUC fixes](https://github.com/igniterealtime/Openfire/pull/2510), they're all green.

guusdk:
I've rebased this PR (and added some changes) to pull in the update to Smack 4.5.0-beta2

src/main/java/org/jivesoftware/smackx/muc/MultiUserChatOwnerAdminListIntegrationTest.java
Fishbowler: My point is that the code would never run, because the check in the constructor will have already thrown the TestNotPossibleException

Fishbowler: Although there is a slight difference in the error conditions (not authorised vs any error)... :thinking:

Fishbowler: Not a hill I'm gonna die on.

Fishbowler (approved):

PR #42, opened 2024-08-15
Add tests for XEP-0045 Multi-User Chat (section 7.5)

guusdk:


guusdk:
The test failure highlights a bug in Openfire: https://igniterealtime.atlassian.net/browse/OF-2858

A fix is proposed in https://github.com/igniterealtime/Openfire/pull/2505

src/main/java/org/jivesoftware/smackx/muc/MultiUserChatOccupantPMIntegrationTest.java
```diff
@@ -0,0 +1,294 @@
...
+                    if (participant.equals(targetMucAddress)) {
+                        ownerSeesTarget.signal();
+                    }
+                }
+            });
+
+            mucAsSeenByTarget.join(nicknameTarget);
+            ownerSeesTarget.waitForResult(timeout);
+
+            conTwo.addStanzaListener(pmListener, new AndFilter(StanzaTypeFilter.MESSAGE, new StanzaIdFilter(randomString)));
```
Fishbowler, line 112: Does this filter also need to validate that the `from` address matches conOne's occupant JID (as opposed to their real JID)?
I don't think it quite warrants a test of its own.

Fishbowler (approved):
Aside from my 1 question above about a potential missing check, LGTM!

src/main/java/org/jivesoftware/smackx/muc/MultiUserChatOccupantPMIntegrationTest.java
guusdk: That's a good idea! I've just pushed a new commit that adds that assertion.

PR #44, opened 2024-08-20
Improve tests for XEP-0045 Multi-User Chat (section 9)

guusdk:
Various tests did not explicitly use an admin user. Even when an owner technically should have all admin privileges, it is cleaner to explicitly execute the system under test with an admin user, when testing admin functionality.


guusdk:
Rebased

Fishbowler (approved):
Ooof. That was big.

PR #45, opened 2024-08-20
2e containerise

guusdk:


entrypoint.sh
```diff
@@ -0,0 +1,86 @@
...
+if [ ! -v ADMINPASSWORD ] && [ -v ADMINUSERNAME ]; then echo "Admin password is not set, but username is. Credentials must be specified as a pair"; exit 1; fi
+
+# If host is local, replace with host.docker.internal
+#if [ "$HOST" == "127.0.0.1" ] || [ "$HOST" == "localhost" ]; then
+#  HOST="host.docker.internal"
+#fi
+
+java \
+  -Dsinttest.service="$SERVICE" \
+  -Dsinttest.host="$HOST" \
```
Fishbowler, line 74: What happens if HOST isn't set?

guusdk:
```
Saving JUnit-compatible XML file with results to /logs/test-results.xml
Saving debug logs in /logs
Exception in thread "main" org.jivesoftware.smack.SmackException$EndpointConnectionException: Could not lookup the following endpoints: RemoteConnectionEndpointLookupFailure(description='DNS lookup exception for .' exception='java.net.UnknownHostException: .: No address associated with hostname')
	at org.jivesoftware.smack.SmackException$EndpointConnectionException.from(SmackException.java:340)
	at org.jivesoftware.smack.tcp.XMPPTCPConnection.connectUsingConfiguration(XMPPTCPConnection.java:680)
	at org.jivesoftware.smack.tcp.XMPPTCPConnection.connectInternal(XMPPTCPConnection.java:863)
	at org.jivesoftware.smack.AbstractXMPPConnection.connect(AbstractXMPPConnection.java:546)
	at org.igniterealtime.smack.inttest.XmppConnectionManager.<init>(XmppConnectionManager.java:189)
	at org.igniterealtime.smack.inttest.SmackIntegrationTestFramework.run(SmackIntegrationTestFramework.java:220)
	at org.igniterealtime.smack.inttest.SmackIntegrationTestFramework.main(SmackIntegrationTestFramework.java:115)
```

guusdk: I'm adding defaults for HOST and SERVICE, just like we do with the GitHub Action. That prevents these.

guusdk:
So... how would users (programmatically) detect that there are test failures? We don't have something like an exit code, do we?

guusdk:
Oh, we _do_ have an exit code. Executing `echo $?` right after running the docker-based tests will give you a non-zero exit code if one or more of the tests fail.

guusdk:
As Dan is traveling and I'm impatient, I'll merge this now and ask for forgiveness later.

PR #50, opened 2024-09-01
Add docker testing to PR CI

Fishbowler:
Add a build-and-test step for container builds as a test of the containerisation.
We expose some logs, but don't keep the container, nor do we publish test results - other jobs deal with the tests themselves.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,7 +9,7 @@ permissions:
   checks: write
 
 jobs:
-  build:
+  build-java:
     name: Build
     runs-on: ubuntu-latest
     strategy:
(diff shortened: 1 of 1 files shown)


Fishbowler:
@guusdk This further reinforces my desire to have an "enabled" as well as a "disabled" switch. I wish this could run one trivial test to prove the structure.

Fishbowler:
Failures appear consistent with what we'd expect right now.

guusdk:
> @guusdk This further reinforces my desire to have an "enabled" as well as a "disabled" switch. I wish this could run one trivial test to prove the structure.

Let's at the very least track that in a dedicated issue.

PR #51, opened 2024-09-01
Podman file & CI

Fishbowler:
`podman` is CLI and API compatible with `docker` in most instances. Important things of note for us:

- podman doesn't prefer Docker Hub, and by default has no unqualified registries to search
- podman _can_ use Docker Hub for dependencies
- podman prefers a Containerfile to a Dockerfile (although it would use both)

Docker and OCI images are almost identical, so we can use a lot of what was done in the Dockerfile for the Containerfile.

This PR:
- Creates a Containerfile that is a duplicate of the Dockerfile with explicit registries for the JDK image
- Adds a CI job that builds the Containerfile using podman (but doesn't test it - that seems superfluous at this stage)

---

Alternative approaches:
- Don't have a Containerfile, and explicitly state the registries in the Dockerfile
- Don't have a Containerfile, and use a [buildah](https://github.com/containers/buildah?tab=readme-ov-file#example) script instead


guusdk:
I have no experience with any of this, so my comments should be easily dismissed in favor of someone's that actually has familiarity with this.

Going with the 'more native' approach (Containerfile over Dockerfile) seems to make sense to me.

I have no idea of Buildah over Containerfile makes sense. From the Buildah readme:

> Because of this and their underlying storage differences, you can not see Podman containers from within Buildah or vice versa.

Is this a full mutual exclusion? If so, does it make sense to provide _both_, rather than just one?

guusdk:
We should probably add documentation to the README for this.

guusdk:
As we discussed in our video call: buildah is probably not needed. It doesn't offer features for our end-users, and we can manage creating the images without buildah perfectly fine.

guusdk (approved):

Fishbowler:
With the image temporarily building for PRs, I was able to get this locally:
image

PR #58, opened 2024-09-02
Fix the actor in several tests that assert admins modifying moderator list

guusdk:
In tests that verify that a user that is an admin can or cannot perform certain actions, the actor should be an admin.

Prior to this commit, the actor that was used for certain tests was an owner (a super-class of 'admin'). This is done to work around a limitation in which only three test users can be done. In certain tests, using an owner negates the purpose of the test (as an owner is allowed to do some things that an admin is not).


Fishbowler (approved):

Fishbowler:
Some of these switches are hard to follow, but they look sensible :+1:

guusdk:
After rebasing this, it turns out that we already did most of the same changes in another PR. :facepalm:

I'm merging what ends up to be a couple of comments only with this PR.

PR #59, opened 2024-09-02
Fix VCard error equality check

guusdk:
When requesting a vCard from a non-existing user, an error must be returned. When requesting a vCard from a user that has no vCard, an error must also be returned.

There are two tests in our code that check the above.

There is a third definition (and we have test for that), that states that both errors must be the same.

When one of the first two tests fail, running the third test is not very useful. This commit skips that test in that case.

### DIFF
--- a/src/main/java/org/igniterealtime/smack/inttest/xep0054/VCardTempIntegrationTest.java
+++ b/src/main/java/org/igniterealtime/smack/inttest/xep0054/VCardTempIntegrationTest.java
@@ -299,7 +299,7 @@ public class VCardTempIntegrationTest extends AbstractSmackIntegrationTest
 
         if (errorNonExistingVCard == null) {
             // If this throws, then the test testForErrorWhenRecipientHasNoVCard() should have failed.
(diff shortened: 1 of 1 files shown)


src/main/java/org/igniterealtime/smack/inttest/xep0054/VCardTempIntegrationTest.java
```diff
@@ -297,13 +297,23 @@ public void testForErrorEquality() throws SmackException.NotConnectedException,
             errorNonExistingVCard = e.getStanzaError();
         }
 
+        if (errorNonExistingVCard == null) {
+            // If this throws, then the test testForErrorWhenRecipientHasNoVCard() should have failed.
+            throw new TestNotPossibleException("Requesting a vCard from an entity did does not have a vCard unexpectedly did not result in an error.");
```
Fishbowler, line 302: Typo
```suggestion
            throw new TestNotPossibleException("Requesting a vCard from an entity that does not have a vCard unexpectedly did not result in an error.");
```

PR #61, opened 2024-09-05
Leave system in predicatable state after SubscriptionTests

guusdk:
When tests have finished executing, they should leave the state of the XMPP users in a consistent state. This prevents tests from having results that are affected by the exection of other tests.

The goal of this commit is to ensure that any change to subscription state that is made in the SubscriptionTests is undone.


guusdk:
I'm hoping that this resolves some inconsistent behavior that has been observed while running the tests against ejabberd.

guusdk:
The one failing test is the one that keeps failing but was fixed in a future version of Openfire.

PR #63, opened 2024-09-07
Add integration tests for XEP-0486 "MUC Avatars"

guusdk:
Sadly the XEP is not using very strict wording. These tests (of which 2 are duplicates) is all that's strictly defined, as far as I can tell.

### DIFF
--- a/src/main/java/org/jivesoftware/smackx/muc/MUCAvatarIntegrationTest.java
+++ b/src/main/java/org/jivesoftware/smackx/muc/MUCAvatarIntegrationTest.java
@@ -1,0 +1,221 @@
+/**
+ * Copyright 2024 Guus der Kinderen
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


Fishbowler:
Something's gone awry with the commit wording here

Fishbowler:
Stole the wording from a previous version of the commit :heavy_check_mark:

src/main/java/org/jivesoftware/smackx/muc/MUCAvatarIntegrationTest.java
```diff
@@ -0,0 +1,216 @@
...
+            vCard.setTo(mucAddress);
+            vCard.setType(IQ.Type.set);
+            vCard.setStanzaId();
+
+            // Execute system under test.
+            conOne.sendIqRequestAndWaitForResponse(vCard);
+
+            // Verify result.
+        } catch (XMPPException.XMPPErrorException e) {
+            fail("Expected '" + conOne.getUser() + "' (an owner) to be able to unpublish an avatar (via vCard) for '" + mucAddress + "' (but the server returned an error).", e);
```
Fishbowler, line 148: Worth checking that it was honoured?
There aren't any tests (in here) for retrieval with no vcard.

guusdk: That's a good idea. I added that now. I messed up the commit a bit and amended it to an unrelated commit that's also in this PR, but at least the change is in _somewhere_ now. :)

Fishbowler (approved):

PR #65, opened 2024-09-26
Update to Smack 4.5.0-beta3-SNAPSHOT

guusdk:
Today's beta3-SNAPSHOT includes a major change in Smack's build system, which also updates its minimum Java requirement from 8 to 11 (we were already using 11), and fixes the bug in the FormTest that was recently introduced.


guusdk:
Replaced by https://github.com/XMPP-Interop-Testing/smack-sint-server-extensions/pull/66

PR #70, opened 2024-10-17
Add enabledTests and enabledSpecifications to the container entrypoint

Fishbowler:
Since SINT already supports these options, this is just a plumbing job.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -92,7 +92,7 @@ jobs:
           -Dsinttest.adminAccountPassword=admin \
           -Dsinttest.enabledConnections=tcp \
           -Dsinttest.dnsResolver=javax \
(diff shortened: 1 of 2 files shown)


Fishbowler:
The logic switch is to stop empty fields for enabledTests and enabledSpecifications from running 0 tests

guusdk:
Should we add documentation (in this repo maybe, but surely the website) and possibly make use of one of the new options in the CI of this repo?

guusdk (approved):
I've modified CI to make use of one of the new flags. That serves as a test for this change, but also makes our CI more robust against new tests that (correctly) flag issues in the server that we're using here to run the tests against.

Documentation should be added to the runners that will be based on this change, not necessarily to this repository itself.

PR #73, opened 2025-03-11
Add tests for XEP-0421: Occupant identifiers for semi-anonymous MUCs

guusdk:


Fishbowler (approved):

Fishbowler:
There's some MUSTs in the XEP that you haven't attempted here.

**Messages and presences MUST NOT contain more then one <occupant-id> element** 

Is this hard to test due to implementation details?

**The occupant identifier MUST be generated such that it is unique. This means that it MUST be sufficiently improbable that one user is able to re-create the occupant identifier of another user.**

This test would be a bit slow, I imagine? Join 100 temporary rooms, add occupant IDs to a set, check the set size?

**The occupant identifier MUST be generated such that it is pseudonymous. This means that it MUST be sufficiently hard to determine the real bare JID of an occupant from its occupant identifier.**

Is a simple test better than no test? Like checking that the localpart doesn't exist in the occupant-id?

**The occupant identifier MUST have a maximum length of 128 characters.**

This one looks easy enough to test, right?


Other than that, should we be testing persistent v temporary rooms? Behaviour isn't specified by the XEP, and the config form doesn't seem to have a default defined either.

guusdk:
> **Messages and presences MUST NOT contain more then one element**
> 
> Is this hard to test due to implementation details?

I'm not sure how to write an explicit test for this. I could implicitly assert in every test that there's only one element (what we currently do is use the _first_ element). I feel like an explicit test should somehow attempt to create a scenario in which more than one element could be present. `testOccupantIdInReflectedMessageReplaced` and `testOccupantIdInBroadcastMessageReplaced` kind of do this. What is a good approach here?

guusdk:
> **The occupant identifier MUST be generated such that it is unique. This means that it MUST be sufficiently improbable that one user is able to re-create the occupant identifier of another user.**
> 
> This test would be a bit slow, I imagine? Join 100 temporary rooms, add occupant IDs to a set, check the set size?

Do you feel that this sufficiently proves anything? 100 rooms with one ID is hardly a valid collision detection, if you ask me. We could add it, but I fear that it is more likely to cause issues (with rate limiting, for example, or with sheer overhead of logged data) than that it'd give any kind of trustworthy reassurance?

guusdk:
> **The occupant identifier MUST be generated such that it is pseudonymous. This means that it MUST be sufficiently hard to determine the real bare JID of an occupant from its occupant identifier.**
> 
> Is a simple test better than no test? Like checking that the localpart doesn't exist in the occupant-id?

That's a good idea. I've added a quick test that fails like this when the occupant-id contains the literal username:

```
• XEP-0421: Anonymous unique occupant identifiers for MUCs, Section 4.1
      "The occupant identifier MUST be generated such that it is pseudonymous. This means that it MUST be sufficiently hard to determine the real bare JID of an occupant from its occupant identifier."
  Failure reason  : Expected the occupant-id value ('smack-inttest-one-lqwvu') for user 'smack-inttest-one-lqwvu@example.org/one-lqwvu' in room 'smack-inttest-xep0421-lqwvu@conference.example.org' to be psuedonymous. However, the local-part of the user's real JID is a literal part of the occupant-id. ==> expected: <false> but was: <true>
  Stanza log file : ./target/logs/OccupantIdIntegrationTest.testOccupantIdDoesntContainUsername (Normal).log
  Test class      : class org.igniterealtime.smack.inttest.xep0421.OccupantIdIntegrationTest
  Test method     : testOccupantIdDoesntContainUsername
```

guusdk:
> **The occupant identifier MUST have a maximum length of 128 characters.**
> 
> This one looks easy enough to test, right?

Indeed. Assuming that generally speaking, the generated occupant-id values will be of a similar composition, testing one will be a good indication if the computation of all occupant-ids falls withing the appropriate range.

I've added a test that fails like this:

```
• XEP-0421: Anonymous unique occupant identifiers for MUCs, Section 4.1
      "The occupant identifier MUST have a maximum length of 128 characters."
  Failure reason  : Expected the occupant-id value for user 'smack-inttest-one-yygf8@example.org/one-yygf8' in room 'smack-inttest-xep0421-yygf8@conference.example.org' to have at most 128 characters. However, value consists of more: 800. Offending occupant-id: 'c6b31a065d1b4cee4e7182898b8c06a878fcdbd9c6b31a065d1b4cee4e7182898b8c06a878fcdbd9c6b31a065d1b4cee4e7182898b8c06a878fcdbd9c6b31a065d1b4cee4e7182898b8c06a878fcdbd9c6b31a065d1b4cee4e7182898b8c06a878fcdbd9c6b31a065d1b4cee4e7182898b8c06a878fcdbd9c6b31a065d1b4cee4e7182898b8c06a878fcdbd9c6b31a065d1b4cee4e7182898b8c06a878fcdbd9c6b31a065d1b4cee4e7182898b8c06a878fcdbd9c6b31a065d1b4cee4e7182898b8c06a878fcdbd9c6b31a065d1b4cee4e7182898b8c06a878fcdbd9c6b31a065d1b4cee4e7182898b8c06a878fcdbd9c6b31a065d1b4cee4e7182898b8c06a878fcdbd9c6b31a065d1b4cee4e7182898b8c06a878fcdbd9c6b31a065d1b4cee4e7182898b8c06a878fcdbd9c6b31a065d1b4cee4e7182898b8c06a878fcdbd9c6b31a065d1b4cee4e7182898b8c06a878fcdbd9c6b31a065d1b4cee4e7182898b8c06a878fcdbd9c6b31a065d1b4cee4e7182898b8c06a878fcdbd9c6b31a065d1b4cee4e7182898b8c06a878fcdbd9' ==> expected: <true> but was: <false>
  Stanza log file : ./target/logs/OccupantIdIntegrationTest.testOccupantIdLength (Normal).log
  Test class      : class org.igniterealtime.smack.inttest.xep0421.OccupantIdIntegrationTest
  Test method     : testOccupantIdLength
```

guusdk:
> Other than that, should we be testing persistent v temporary rooms? Behaviour isn't specified by the XEP, and the config form doesn't seem to have a default defined either.

Why do you ask? Do you think this particular room configuration more relevant in this context than any other room configuration variation?

Fishbowler:
> > **Messages and presences MUST NOT contain more then one element**
> > Is this hard to test due to implementation details?
> 
> I'm not sure how to write an explicit test for this. I could implicitly assert in every test that there's only one element (what we currently do is use the first element).

Yeah, I spotted that in Smack. We could use `getExtensions` instead of `getExtension` for the explicit test.

>  I feel like an explicit test should somehow attempt to create a scenario in which more than one element could be present. testOccupantIdInReflectedMessageReplaced and testOccupantIdInBroadcastMessageReplaced kind of do this. 

That feels like a legit place to do it, but we should probably make a copy of the test and make the test explicit.




> > **The occupant identifier MUST be generated such that it is unique. This means that it MUST be sufficiently improbable that one user is able to re-create the occupant identifier of another user.**
> > This test would be a bit slow, I imagine? Join 100 temporary rooms, add occupant IDs to a set, check the set size?
> 
> Do you feel that this sufficiently proves anything?

No, you're right. There's no real way to prove randomness in an integration test like this.




> > **The occupant identifier MUST have a maximum length of 128 characters.**
> > This one looks easy enough to test, right?
> 
> Indeed. Assuming that generally speaking, the generated occupant-id values will be of a similar composition, testing one will be a good indication if the computation of all occupant-ids falls withing the appropriate range.

Agree.



> > Other than that, should we be testing persistent v temporary rooms? Behaviour isn't specified by the XEP, and the config form doesn't seem to have a default defined either.
> 
> Why do you ask? Do you think this particular room configuration more relevant in this context than any other room configuration variation?

Only that variation is variation.
The only real risk I can see is "a MUC service SHOULD generate the identifier such that the occupant identifier of a user in one room of the service does not match the occupant identifier of the same user in another room of the same service". We've hit problems with same-named MUCs before in Openfire. It stands to reason that 2 temporary rooms with the same name shouldn't have the same occupant-id either. Although, come to think of it, that applies to 2 persistent rooms too.... :thinking:

guusdk:
> > > **Messages and presences MUST NOT contain more then one element**
> > > Is this hard to test due to implementation details?
> > 
> > 
> > I'm not sure how to write an explicit test for this. I could implicitly assert in every test that there's only one element (what we currently do is use the first element).
> 
> Yeah, I spotted that in Smack. We could use `getExtensions` instead of `getExtension` for the explicit test.

That's not what I was getting at. We can indeed use that to add an assertion that there's not more than one of those extensions in existing tests, but that wouldn't be a new/standalone test, annotated with something like:

    @SmackIntegrationTest(section = "4", quote = "Messages and presences MUST NOT contain more then one <occupant-id> element."

I guess I could duplicate a couple of tests to get this explicitly checked out. Alternatively, I add the _assertion_ to each preexisting test. That would not give us dedicated tests though (I like having tests dedicated to verifying exactly one thing).

guusdk:
> > > **Messages and presences MUST NOT contain more then one element**
> > > Is this hard to test due to implementation details?
> > 
> > 
> > I'm not sure how to write an explicit test for this. I could implicitly assert in every test that there's only one element (what we currently do is use the first element).
> 
> Yeah, I spotted that in Smack. We could use `getExtensions` instead of `getExtension` for the explicit test.
> 
> > I feel like an explicit test should somehow attempt to create a scenario in which more than one element could be present. testOccupantIdInReflectedMessageReplaced and testOccupantIdInBroadcastMessageReplaced kind of do this.
> 
> That feels like a legit place to do it, but we should probably make a copy of the test and make the test explicit.

I've now added a commit that adds a bunch of explicit tests for "Messages and presences MUST NOT contain more then one element". They're largely a copy of preexisting tests. I have refactored the code to reduce code duplication (see commit message for details).

> > > Other than that, should we be testing persistent v temporary rooms? Behaviour isn't specified by the XEP, and the config form doesn't seem to have a default defined either.
> > >
> > >
> > Why do you ask? Do you think this particular room configuration more relevant in this context than any other room configuration variation?
> >
> >
> Only that variation is variation.

It's one variation of various others (password-locked rooms, members-only rooms, room logging enabled, etc). I'm not _against_ creating tests for these variations, but won't do that now, because:
- I'm not sure how to do this without either creating a lot of code duplication, or adding a lot more complexity.
- I'm not sure if these variations are required to be supported by the XEP. That makes for more complexity.
- I don't think Smack support creating persistent vs. non-persistent rooms (at least not in `MucConfigFormManager` it appears).

> The only real risk I can see is "a MUC service SHOULD generate the identifier such that the occupant identifier

I stopped reading after SHOULD ;) (if I read this correctly, your worry is something that's not mandated - as we previously agreed, we'll focus on mandated specifications only, to avoid rabbit holes that involve edge-cases).

PR #74, opened 2025-03-17
Ad-hoc tests specific to Openfire

Fishbowler:
WIP, includes / depends on #4


Fishbowler:
This needs rebasing when #4 is merged

PR #76, opened 2025-03-28
Release 1.5.0

Fishbowler:
Release the value of enabledTests and enabledSpecifications that was developed and merged last year


Fishbowler:
Will beg forgiveness....

PR #77, opened 2025-04-11
Add tests for XEP-0433: Extended Channel Search

guusdk:


src/main/java/org/igniterealtime/smack/inttest/xep0433/ExtendedChannelSearchAllIntegrationTest.java
```diff
@@ -0,0 +1,243 @@
...
+        final Form form = new Form(DataForm.from(response));
+
+        final ExtendedChannelSearchForm searchRequest = new ExtendedChannelSearchForm();
+        searchRequest.setType(IQ.Type.get);
+        searchRequest.setTo(searchService);
+
+        // Attempt to set the 'all' field
+        final FillableForm fillableForm = form.getFillableForm();
+        for (final FormField field : fillableForm.getDataForm().getFields()) {
+            if (field.getFieldName().equalsIgnoreCase("all")) {
```
Fishbowler, line 208: Another check for 'all' support

src/main/java/org/igniterealtime/smack/inttest/xep0433/ExtendedChannelSearchAllIntegrationTest.java
```diff
@@ -0,0 +1,243 @@
...
+                }
+                if (stanzaError.getType().equals(StanzaError.Type.AUTH) && stanzaError.getCondition().equals(StanzaError.Condition.forbidden)) {
+                    throw new TestNotPossibleException("Unable to execute search, as the service has does not allow this test-user to perform queries for the full result set.", e);
+                }
+            }
+            throw e;
+        }
+
+        // Verify result.
+        assertNotNull(searchResponse, "The search request issued by '" + conOne.getUser() + "' did not result in a response from '" + searchService + "'." );
```
Fishbowler, line 241: Assert that the response contains items too?

src/main/java/org/igniterealtime/smack/inttest/xep0433/ExtendedChannelSearchSortIntegrationTest.java
```diff
@@ -0,0 +1,279 @@
...
+        if (field == null) {
+            throw new TestNotPossibleException("The service does not support the 'all' search form field.");
+        }
+    }
+
+    @BeforeClass
+    public void setUp() throws TestNotPossibleException, XMPPException.XMPPErrorException, SmackException.NotConnectedException, SmackException.NoResponseException, InterruptedException
+    {
+        // Create a number of rooms that will act as search results.
+        final MultiUserChatManager mucManager = MultiUserChatManager.getInstanceFor(connection);
```
Fishbowler, line 90: Why `connection` rather than the more explicit `conOne`?

src/main/java/org/igniterealtime/smack/inttest/xep0433/ExtendedChannelSearchSortIntegrationTest.java
```diff
@@ -0,0 +1,279 @@
...
+                mucManager.getMultiUserChat(mucAddress).create(Resourcepart.from("test-user-one")).getConfigFormManager().setRoomName("Test Room " + i).submitConfigurationForm();
+                if (i % 2 == 0) {
+                    mucManagerTwo.getMultiUserChat(mucAddress).join(Resourcepart.from("test-user-two"));
+                }
+                if (i % 3 == 0) {
+                    mucManagerThree.getMultiUserChat(mucAddress).join(Resourcepart.from("test-user-three"));
+                }
+            }
+        } catch (Exception e) {
+            throw new TestNotPossibleException("Unable to create or popluate MUC room.", e);
```
Fishbowler, line 108: Typo
```suggestion
            throw new TestNotPossibleException("Unable to create or populate MUC room.", e);
```

src/main/java/org/igniterealtime/smack/inttest/xep0433/ExtendedChannelSearchSortIntegrationTest.java
```diff
@@ -0,0 +1,279 @@
...
+        searchRequest.addExtension(fillableForm.getDataFormToSubmit());
+
+        try {
+            // Execute system under test.
+            final ExtendedChannelResult result = conOne.sendIqRequestAndWaitForResponse(searchRequest);
+
+            // Verify result.
+            if (result.getItems().size() < 3) {
+                throw new TestNotPossibleException("The service did not return enough results for the implementation to be able to assert that they're consistently ordered.");
+            }
```
Fishbowler, line 176: Given we're creating data in the setup, can't we guarantee some number of results?
The only barriers I can see:
* RSM defaults to < 3 (so we'd have to page)
* The vague "the service may process the keyword string in implementation-defined ways"

Whether or not we choose to page, I'd be slightly inclined towards assuming a sane implementation and assuming that a search for "smack-inttest-xep0433" or for "Test Room" would return sufficient results to test with.

I'm worried that this check is more likely to hide an issue than it is to avoid a false negative. We could let a tester disable the test (or submit a PR) if their implementation really doesn't work in what we've assumed is a predictable manner.

src/main/java/org/igniterealtime/smack/inttest/xep0433/ExtendedChannelSearchSortIntegrationTest.java
```diff
@@ -0,0 +1,279 @@
...
+            final List<String> addresses = result.getItems().stream().map(item -> item.address).collect(Collectors.toList());
+            assertTrue(isSorted(addresses, true) || isSorted(addresses, false), "Expected the items in the response to the search query to be ordered by address (but they were not).");
```
Fishbowler, line 178: I think that it's safe to assume that when a XEP says strings are sorted, that they're sorted ascending.

Fishbowler (changes requested):
Mostly good, but 2 things stuck out:

* There are some places where I think you're playing too safe on assumptions, and risk hiding problems that the tests could expose
* This puts a bunch of effort into creating data in setUp methods, but doesn't try to assert on anything correct appearing in the Items. The only test for Item structure is implied by not catching a parsing exception.

src/main/java/org/igniterealtime/smack/inttest/xep0433/ExtendedChannelSearchIntegrationTest.java
```diff
@@ -0,0 +1,509 @@
...
+        // Destroy the rooms that were used as search results.
+        final MultiUserChatManager mucManager = MultiUserChatManager.getInstanceFor(connection);
+        final DomainBareJid mucDomain = mucManager.getMucServiceDomains().stream().findFirst().orElseThrow(() -> new IllegalStateException("Unable to find a MUC service domain"));
+        for (int i = 1; i <= ROOMS_AMOUNT; i++) {
+            String roomNameLocal = String.join("-", ROOM_NAME_PREFIX, testRunId, Integer.toString(i));
+            EntityBareJid mucAddress = JidCreate.entityBareFrom(Localpart.from(roomNameLocal), mucDomain);
+            mucManager.getMultiUserChat(mucAddress).destroy();
+        }
+    }
+    @SmackIntegrationTest(section = "4.2", quote = "the Searcher MAY [...] request the search form from the Search Service.")
```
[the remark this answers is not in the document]
guusdk: Yeah, if the client _MAY_ request it, then the server _MUST_ be able to process the request with a non-error response, was my thinking.

I've been providing the author of the XEP (@horazont) with a list of feedback (which perhaps I should've done on a list, but I've done that in a direct message) on the XEP, based on me writing these tests. An improvement around this was in that list. I think that would cover your concern, too. I've worded it like this:

> I'd like to see a MUST in 4.2 / 4.2.1 that describes that the service to be searched MUST support the retrieval of a search form. That removes ambiguity, I think. I prefer the clarity of the wording in the comparable section 2.1 of XEP-0055: https://xmpp.org/extensions/xep-0055.html#usecases-search That section describes basically the same functionality requesting a form), but it's more explicit.

src/main/java/org/igniterealtime/smack/inttest/xep0433/ExtendedChannelSearchIntegrationTest.java
```diff
@@ -0,0 +1,509 @@
...
+    final DomainBareJid searchService;
+
+    public ExtendedChannelSearchIntegrationTest(SmackIntegrationTestEnvironment environment) throws XMPPException.XMPPErrorException, SmackException.NotConnectedException, SmackException.NoResponseException, InterruptedException, TestNotPossibleException
+    {
+        super(environment);
+
+        searchService = ServiceDiscoveryManager.getInstanceFor(connection).findService("urn:xmpp:channel-search:0:search", false);
+        if (searchService == null) {
+            throw new TestNotPossibleException("Unable to find any service on domain that supports XEP-0433: Extended Channel Search.");
+        }
```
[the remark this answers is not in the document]
guusdk: What would be a reason to put this in `setUp()`? I have no strong preference, and I'd be happy to change things.

My rationale for splitting up the bits as they are now:
- Things that must be cleaned up again after running the tests (the rooms created for the test) require a `@AfterClass` method. I prefer to keep things symmetric (which generally helps keeping code clean, I've found), so anything teared down in `@AfterClass` is set up in a method annotated with `@BeforeClass`
- Basic initialization of helper services (like `searchService`) is done in the constructor, which populates a class member. I don't think this can be done in a static method like `@BeforeClass`, as the class instance doesn't exist yet when that's invoked
- Throwing `TestNotPossibleException` typically happens in the constructor, I think. I don't know if we _can_ throw it in an `@BeforeClass` annotated method. I'd not be against doing something like that (but I'm also not sure if there's a benefit of doing it there, versus in the constructor).

src/main/java/org/igniterealtime/smack/inttest/xep0433/ExtendedChannelSearchIntegrationTest.java
```diff
@@ -0,0 +1,509 @@
...
+        // Setup test fixture.
+        final ExtendedChannelSearchForm formRequest = new ExtendedChannelSearchForm();
+        formRequest.setType(IQ.Type.get);
+        formRequest.setTo(searchService);
+        final IQ response = conOne.sendIqRequestAndWaitForResponse(formRequest);
+        final Form form = new Form(DataForm.from(response));
+
+        final ExtendedChannelSearchForm searchRequest = new ExtendedChannelSearchForm();
+        searchRequest.setType(IQ.Type.get);
+        searchRequest.setTo(searchService);
```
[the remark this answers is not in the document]
guusdk: For me, this is logically where it should be placed: the first request obtains the form. Then, a second request is created, that's then populated with a form from the response of the first request. If this bothers you, I'd be happy to turn things around though. :)

src/main/java/org/igniterealtime/smack/inttest/xep0433/ExtendedChannelSearchIntegrationTest.java
```diff
@@ -0,0 +1,509 @@
...
+        } catch (XMPPException.XMPPErrorException e) {
+            if (e.getStanzaError().getCondition().equals(StanzaError.Condition.resource_constraint)) {
+                throw new TestNotPossibleException("Unable to execute search, as the service is rejecting the request due to rate limiting", e);
+            }
+            throw e;
+        }
+
+        // Verify result.
+        assertNotNull(searchResponse, "The search request issued by '" + conOne.getUser() + "' did not result in a response from '" + searchService + "'." );
+        // We cannot assert this, as the RMS XEP explicitly states that no RMS element is present when the total result set contains 0 items). assertTrue(searchResponse.hasExtension(RSMSet.class), "Expected the response from '" + searchService + "' to the search request issued by '" + conOne.getUser() + "' to contain a Result Set Management extension (but it did not).");
```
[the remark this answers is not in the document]
guusdk: I'm not sure that we can guarantee that the search yields a result. The primary purpose of this test is to verify that a client can provide an RSM element in their request, which I believe the current implementation already does (by ensuring a no-error response).

If we add that assertion for an RSM element in the response, and that check fails during a test, then we've got a discussion on our hands if that's a false positive. My reasoning is to avoid that - at least in context of _this_ test.

If we'd like to assert that an RSM element is present in the result, then maybe we should add a test like this, but in its verificatino ph ase evaluate if the response has at least one item, and then make this assertion (or throw a TestNotPossible if the response didn't have any items).

src/main/java/org/igniterealtime/smack/inttest/xep0433/ExtendedChannelSearchIntegrationTest.java
```diff
@@ -0,0 +1,509 @@
...
+            if (e.getStanzaError().getCondition().equals(StanzaError.Condition.item_not_found)) {
+                fail("Expected the search service '" + searchService + "' to return an empty result to '" + conOne.getUser() + "' when the search does not yield results, but the service returned an 'item-not-found' error.");
+            }
+            return;
+        }
+
+        assertNotNull(searchResponse, "The search request issued by '" + conOne.getUser() + "' did not result in a response from '" + searchService + "'.");
+        if (!searchResponse.getItems().isEmpty()) {
+            throw new TestNotPossibleException("A non-empty response was returned to a search request that was not intended to return any items in the response.");
+        }
```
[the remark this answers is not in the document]
guusdk: Your reasoning assumes that the search query includes that specific 'needle'. That is 'only' true if the request was actually made with that value in the `q` field. That field is not a _required_ field. If the request does not include that field, then chances are that the request is being interpreted as a request for all rooms. This is why I chose to throw a `TestNotPossibleException` rather than failing the test.

I do think the test can easily be improved though. It can track _if_ the `q` field was set, and use that to see if it should fail the test, or throw a `TestNotPossibleException`. What do you think?

src/main/java/org/igniterealtime/smack/inttest/xep0433/ExtendedChannelSearchSortIntegrationTest.java
guusdk: Eh, probably laziness or copy/paste from the setup/constructor code. I agree that `conOne` is better. I'll change that.

src/main/java/org/igniterealtime/smack/inttest/xep0433/ExtendedChannelSearchSortIntegrationTest.java
guusdk: Shouldn't that be the subject of a different test then? I prefer to keep the thing that's being tested (very) concise. This test asserts if the ordering of results is consistent. For that, you need any three results. If there's a problem that's preventing three results to be included in the response, that aught to be the subject of a different test, right?

src/main/java/org/igniterealtime/smack/inttest/xep0433/ExtendedChannelSearchSortIntegrationTest.java
guusdk: Should we _fail_ the test if it's sorted the other way around for some reason? On one hand, I can see that as a bug, on the other hand I wonder if there's things in RSM that let you page backwards, somehow...

src/main/java/org/igniterealtime/smack/inttest/xep0433/ExtendedChannelSearchIntegrationTest.java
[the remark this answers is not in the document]
Fishbowler: Fair. My only rationale was "this is a setup activity".

src/main/java/org/igniterealtime/smack/inttest/xep0433/ExtendedChannelSearchIntegrationTest.java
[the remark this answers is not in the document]
Fishbowler: It was confusing on first top-to-bottom read. I expected to request the form, fill the form, send the form, with the creation of the second request being part of that last step. When I was reading, the second request didn't make sense, since it was instantiated then not used for a while whilst other things were being done.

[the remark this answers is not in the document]
Fishbowler: But it's very minor and would require changing in a bunch of places. Only change it if you're already changing a bunch of other stuff too.

src/main/java/org/igniterealtime/smack/inttest/xep0433/ExtendedChannelSearchIntegrationTest.java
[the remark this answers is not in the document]
Fishbowler: That makes sense. The alternative is a duplicate with failure vs TestNotPossible, and that feels overly verbose.

src/main/java/org/igniterealtime/smack/inttest/xep0433/ExtendedChannelSearchAllIntegrationTest.java
```diff
@@ -0,0 +1,243 @@
...
+        final Form form = new Form(DataForm.from(response));
+
+        final ExtendedChannelSearchForm searchRequest = new ExtendedChannelSearchForm();
+        searchRequest.setType(IQ.Type.get);
+        searchRequest.setTo(searchService);
+
+        // Attempt to set the 'all' field
+        final FillableForm fillableForm = form.getFillableForm();
+        for (final FormField field : fillableForm.getDataForm().getFields()) {
+            if (field.getFieldName().equalsIgnoreCase("all")) {
```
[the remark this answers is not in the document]
Fishbowler: I had meant the `all` field, because I'd forgotten the constructor check. I'm glad that my error led somewhere useful all the same!

src/main/java/org/igniterealtime/smack/inttest/xep0433/ExtendedChannelSearchAllIntegrationTest.java
guusdk: I think that is covered by the test in the constructor, but it doesn't hurt to have an explicit test in this method too. It is, after all, a different request that for some reason could not include that form field even if it was present in the constructor's response.

src/main/java/org/igniterealtime/smack/inttest/xep0433/ExtendedChannelSearchAllIntegrationTest.java
guusdk: Agreed. I've added an assertion for the items in the response to be non-empty.

src/main/java/org/igniterealtime/smack/inttest/xep0433/ExtendedChannelSearchSortIntegrationTest.java
Fishbowler: RSM will let you page backwards but won't let you reverse the order, so if you've got items 11-15 and page backwards, you ~will~ should get items 6-10 (in that order). 

In short: Yes, fail the test.

src/main/java/org/igniterealtime/smack/inttest/xep0433/ExtendedChannelSearchIntegrationTest.java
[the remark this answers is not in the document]
guusdk: I've now added an explicit check to see if `q` was used. If it was and the result is non-empty, then I'm failing this test now.

src/main/java/org/igniterealtime/smack/inttest/xep0433/ExtendedChannelSearchAllIntegrationTest.java
```diff
@@ -0,0 +1,243 @@
...
+                throw new TestNotPossibleException("Unable to execute search, as the service is rejecting the request due to rate limiting", e);
+            }
+            if (stanzaError.getExtension("full-set-retrieval-rejected", "urn:xmpp:channel-search:0:error") != null) {
+                if (stanzaError.getType().equals(StanzaError.Type.CANCEL) && stanzaError.getCondition().equals(StanzaError.Condition.not_allowed)) {
+                    throw new TestNotPossibleException("Unable to execute search, as the service has generally disabled the queries for the full result set.", e);
+                }
+                if (stanzaError.getType().equals(StanzaError.Type.AUTH) && stanzaError.getCondition().equals(StanzaError.Condition.forbidden)) {
+                    throw new TestNotPossibleException("Unable to execute search, as the service has does not allow this test-user to perform queries for the full result set.", e);
+                }
+            }
```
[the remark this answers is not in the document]
guusdk: Isn't throwing the exception (and thus make the test error out, rather than fail) more appropriate in that case?

src/main/java/org/igniterealtime/smack/inttest/xep0433/ExtendedChannelSearchIntegrationTest.java
```diff
@@ -102,6 +102,7 @@ public void tearDown() throws XMPPException.XMPPErrorException, SmackException.N
             mucManager.getMultiUserChat(mucAddress).destroy();
         }
     }
+
```
Fishbowler, line 105: :heart: 
This mssing break bothered me twice, but I couldn't bring myself to nitpick that deeply :laughing:

Fishbowler:
All the changes look good. We're down to additional tests:
* Should we add a test to see if RSM works?
* Should we add a positive test to see if `q` works?
* Should we add a test to evaluate the structure of items?
* Should we add a test for the correctness of items?

guusdk:
>     * Should we add a test to see if RSM works?

Possibly, but that feels like a rabbit hole. Are we going to test all of RSM's feature against any XEP that is RSM-able? What is good middle-ground here? Can we somehow make things re-usable?

>     * Should we add a positive test to see if `q` works?

Yes, that'd be good.

>     * Should we add a test to evaluate the structure of items?

There is only so much that Smack lets us do explicitly. Smack is set up to be very specific, and fails-fast when data cannot be serialized in corresponding extensions. With that, we have some implicit tests (that cause tests to error out rather than fail) when items do not have the correct syntax.

>     * Should we add a test for the correctness of items?

How do you mean? If you're thinking of checking for the exact content of results: I'm fearful that server-sided implementations will use caching, which may make changes to rooms not pop-up immediately.

guusdk:
I've now added a positive `q` test.

Fishbowler:
>> Should we add a test to see if RSM works?

> Possibly, but that feels like a rabbit hole. Are we going to test all of RSM's feature against any XEP that is RSM-able? What is good middle-ground here? Can we somehow make things re-usable?

On second thoughts, lets not. The XEP in its current form is a bit conflicted on whether a server MUST or MAY support RSM at all.



>> Should we add a test to evaluate the structure of items?

> There is only so much that Smack lets us do explicitly. Smack is set up to be very specific, and fails-fast when data cannot be serialized in corresponding extensions. With that, we have some implicit tests (that cause tests to error out rather than fail) when items do not have the correct syntax.

Our client implementation ignores fields it doesn't understand (as per the XEP). If we received a result of the right namespace, containing an item that was missing fields that are expected to be present (like "service-type"), would that be an error?



>> Should we add a test for the correctness of items? 

> How do you mean? If you're thinking of checking for the exact content of results: I'm fearful that server-sided implementations will use caching, which may make changes to rooms not pop-up immediately.

The XEP says that "The number of occupants may be stale by an undefined amount of time." - it's not explicit, but implies that other fields shouldn't be stale, no?

guusdk:
> > > Should we add a test to evaluate the structure of items?
> 
> > There is only so much that Smack lets us do explicitly. Smack is set up to be very specific, and fails-fast when data cannot be serialized in corresponding extensions. With that, we have some implicit tests (that cause tests to error out rather than fail) when items do not have the correct syntax.
> 
> Our client implementation ignores fields it doesn't understand (as per the XEP). If we received a result of the right namespace, containing an item that was missing fields that are expected to be present (like "service-type"), would that be an error?

I think that's arguable, at best, but maybe even false. There's this note:

> Any child element may be omitted by a Search Service if the data is not available for any or all rooms.

So I don't think we can explicitly check for absence of elements.

guusdk:
I have bundled lot of the questions and suggestions for the XEP author in an email, which I've sent the the XSF's "standards" mailing list: https://mail.jabber.org/hyperkitty/list/standards@xmpp.org/thread/XTO3CO3G4NOTE67UZ43L57QKTOCLPMTO/ Feel free to add more feedback there.

Fishbowler:
Rebased

Fishbowler:
Tests pass on my machine.

PR #78, opened 2025-04-14
Additional test for XEP-0421: check that _all_ client-provided occupa…

guusdk:
…nt-ids are removed


Fishbowler (approved):
Good addition!

PR #79, opened 2025-04-22
Add tests for XEP-0410: MUC Self-Ping (Schrödinger's Chat)

guusdk:
There's only so much server-sided functionality that can be tested I think, but I've added a couple of those.

### DIFF
--- a/src/main/java/org/igniterealtime/smack/inttest/xep0410/MucSelfPingIntegrationTest.java
+++ b/src/main/java/org/igniterealtime/smack/inttest/xep0410/MucSelfPingIntegrationTest.java
@@ -218,7 +218,7 @@ public class MucSelfPingIntegrationTest extends AbstractSmackIntegrationTest
             });
 
             final AtomicBoolean wasInvokedB = new AtomicBoolean(false);
-            conOne.registerIQRequestHandler(new IQRequestHandler()
+            conOneSecondary.registerIQRequestHandler(new IQRequestHandler()
             {
                 @Override
                 public IQ handleIQRequest(IQ iqRequest) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/igniterealtime/smack/inttest/xep0410/MucSelfPingIntegrationTest.java
```diff
@@ -0,0 +1,290 @@
...
+
+            // Verify results
+            assertEquals(IQ.Type.result, response.getType(), "Expected a non-error response to the ping request that '" + conOne.getUser() + "' sent to its own nickname in room '" + addressInRoom + "' (but an error response was received.");
+            assertFalse(wasInvoked.get(), "After sending a 'self-ping' to its own nickname in room '" + addressInRoom + "' the server should have processed the request on behalf of the user. Instead, '" + conOne.getUser() + "' received the ping request from the server.");
+        } finally {
+            removeRoom();
+            if (oldPingHandler != null) {
+                conOne.registerIQRequestHandler(oldPingHandler);
+            }
+        }
```
Fishbowler, line 162: Does this need to unregister the new IQRequestHandler too?

Fishbowler (approved):
One comment about tidying up, which would apply to another test too if it's valid. Otherwise, LGTM.

There's some interesting stuff in here were we to ever start writing client tests :)

src/main/java/org/igniterealtime/smack/inttest/xep0410/MucSelfPingIntegrationTest.java
guusdk: I've now added two commits that apply these changes.

Fishbowler (approved):

PR #83, opened 2025-06-17
Add tests for RFC 6121 Section 8.5.2.1.2 'Server Rules for Processing XML Stanzas'

guusdk:
Integration tests that verify that behavior defined in section 8.5.2.1.2 "Local User / localpart@domainpart / Available or Connected Resources / Presence" of section 8 "Server Rules for Processing XML Stanzas" of RFC6121.

The tests here only partially cover this paragraph, as the paragraph references other paragraphs for most of its definition. The implementation added in this commit tests the processing of presence stanzas with no type or of type 'unavailable' only. All other types are/should be subject of tests under other sections (Section 3 and Section 4.3), to which section 8.5.2.1.2 references (and which have a much more detailed definition of the desired functionality).


guusdk:
The implementation of these tests resemble those in https://github.com/XMPP-Interop-Testing/smack-sint-server-extensions/pull/82, but are simpler: as the tests only deal with intended recipient resources that are always expected to receive stanzas, a lot of the complexity that was needed to test if _certain_ resources would receive stanzas (while others would not), isn't needed.

PR #84, opened 2025-06-18
Add tests for RFC 6121 Section 8.5.2.1.3 'Server Rules for Processing XML Stanzas'

guusdk:
Integration tests that verify that behavior defined in section 8.5.2.1.3 "Local User / localpart@domainpart / Available or Connected Resources / IQ" of section 8 "Server Rules for Processing XML Stanzas" of RFC6121.


guusdk:
I have force-pushed an amended fix.

PR #90, opened 2025-07-10
Add tests for RFC 6121 Section 8.5.3.2.2 'Server Rules for Processing XML Stanzas'

guusdk:
Integration tests that verify that behavior defined in section 8.5.3.2.2 "Local User / localpart@domainpart/resourcepart / No Resource Matches / Presence" of section 8 "Server Rules for Processing XML Stanzas" of RFC6121.

Note: this _does not_ tests the processing of presence stanzas with of type 'subscribe' or 'probe'. Those types are subject of tests under other sections (Section 3 and Section 4.3), to which section 8.5.3.2.2 references (and which have a much more detailed definition of the desired functionality).

The implementation of this test uses a utility class that is introduced by https://github.com/XMPP-Interop-Testing/smack-sint-server-extensions/pull/83 . The build will fail until it is rebased onto a branch that has that.

### DIFF
--- a/src/main/java/org/igniterealtime/smack/inttest/rfc6121/section8/RFC6121Section8_5_3_2_2_PresenceIntegrationTest.java
+++ b/src/main/java/org/igniterealtime/smack/inttest/rfc6121/section8/RFC6121Section8_5_3_2_2_PresenceIntegrationTest.java
@@ -1,0 +1,639 @@
+/**
+ * Copyright 2025 Guus der Kinderen
+ *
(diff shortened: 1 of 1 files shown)


Fishbowler:
Ah, but not on #83, since that's not in main yet :facepalm:

Fishbowler (approved):

Fishbowler (approved):
This all looks sensible, but I've not yet seen em run.

guusdk:
This only needed another rebase to make CI happy!

PR #96, opened 2025-07-31
Tests should not assume a default role when joining a MUC

### LINKED ISSUE #95: OccupantIdIntegrationTest should not assume default role
The OccupantIdIntegrationTest creates a moderated room, and assumes that new occupants are automatically given the `visitor` role. This isn't necessarily the case. For example, ejabberd has the configuration option shown below, that causes new occupants to automatically have the `participant` role.

> members_by_default: true | false
> The occupants that enter the room are participants by default, so they have "voice". The default value is true.

This breaks the assumption in the test, which then causes the test to time-out as it tries to change the role of the user to a role that it already has.

guusdk:
When joining a MUC, it cannot be assumed that every server hands out the same default role. For example, ejabberd will, by default, make new occupants 'participants' in moderated rooms, while Openfire does not.

This commit adjusts tests that depend on having a specific role in the test fixture. They now explicitly check if the role on join matches the desired role (and applies a correction if that's not the case).

fixes #95

### DIFF
--- a/src/main/java/org/igniterealtime/smack/inttest/xep0421/OccupantIdIntegrationTest.java
+++ b/src/main/java/org/igniterealtime/smack/inttest/xep0421/OccupantIdIntegrationTest.java
@@ -878,7 +878,20 @@ public class OccupantIdIntegrationTest extends AbstractSmackIntegrationTest
         final Resourcepart nickname = Resourcepart.from("test-user");
 
         try {
-            room.join(nickname);
+            final Presence reflectedJoinPresence = room.join(nickname);
+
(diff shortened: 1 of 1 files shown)


guusdk:
I wonder if it makes sense to apply a similar change for MUC affiliations.

Fishbowler:
> I wonder if it makes sense to apply a similar change for MUC affiliations.

Avoiding assumptions is good. XEP '45 doesn't require a server to implement Admin, Member or Outcast, so we should definitely check that too.

Have created #97 to (loosely) describe the problem.

PR #105, opened 2025-08-21
Improve concurrency when dealing with Roster

### LINKED ISSUE #104: Improve concurrency when dealing with Roster
I suspect that Smack asynchronously processes roster updates, which results in infrequent failures to obtain a roster item through `Roster.getEntry()` for a target that was known to have been added.

guusdk:
I suspect that Smack asynchronously processes roster updates, which results in infrequent failures to obtain a roster item through Roster.getEntry() for a target that was known to have been added.

In this commit, specific bits of code listen for an event emitted by the Roster instance that is expected to be updated, rather than for a stanza that arrives (that implies that the roster is updated).

This is expected to reduce occasional failure of the affected tests.

fixes #104

### DIFF
--- a/src/main/java/org/igniterealtime/smack/inttest/rfc6121/section2/RFC6121Section2_1_SyntaxIntegrationTest.java
+++ b/src/main/java/org/igniterealtime/smack/inttest/rfc6121/section2/RFC6121Section2_1_SyntaxIntegrationTest.java
@@ -5,18 +5,25 @@ import org.igniterealtime.smack.inttest.SmackIntegrationTestEnvironment;
 import org.igniterealtime.smack.inttest.TestNotPossibleException;
 import org.igniterealtime.smack.inttest.annotations.SmackIntegrationTest;
 import org.igniterealtime.smack.inttest.annotations.SpecificationReference;
+import org.igniterealtime.smack.inttest.util.ResultSyncPoint;
 import org.igniterealtime.smack.inttest.util.SimpleResultSyncPoint;
 import org.jivesoftware.smack.*;
 import org.jivesoftware.smack.packet.*;
 import org.jivesoftware.smack.parsing.ParsingExceptionCallback;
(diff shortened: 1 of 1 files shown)


src/main/java/org/igniterealtime/smack/inttest/rfc6121/section2/RFC6121Section2_1_SyntaxIntegrationTest.java
```diff
@@ -61,22 +68,33 @@ protected IQChildElementXmlStringBuilder getIQChildElementBuilder(IQChildElement
...
         // Execute system under test
         try {
-            final IQ result = conOne.sendIqRequestAndWaitForResponse(request);
+            final IQ result = conOne.sendIqRequestAndWaitForResponse(request); // After receiving the result, the expected roster push can still be being processed by Smack (possibly even if it was received earlier)! Don't expect roster.getEntry() to reliably result the entry that was added. Use a roster event listener instead.
 
             // Verify result
             assertEquals(IQ.Type.result, result.getType(), "Unexpected response type received by '" + conOne.getUser() + "' after it sent a Roster Set that included a subscription attribute with a value of 'both'. It was expected that the server would ignore this value.");
-
-            final RosterEntry entry = rosterOne.getEntry(target);
+            final RosterEntry entry = assertResult(targetFoundOnRoster, "Expected the roster of '" + conOne.getUser() + "' to have received a push for '" + target + "', even if it was added with an invalid subscription attribute. The expectation is for the server to ignore this value. However, the roster entry was not found.");
```
Fishbowler, line 90: If the problem is that the roster push is still in motion, do we not need to `targetFoundOnRoster.waitForResult(timeout)` too?

Fishbowler: Missed the `assertResult` 🤦

Fishbowler (approved):

PR #111, opened 2025-08-22
Fix #110: Inconsistent results for RosterIntegrationTest

### LINKED ISSUE #110: Flapping RosterIntegrationTest
The RosterIntegrationTest frequently, but not always fails when running the test suite. It does not fail when executed in isolation.

After extensive debugging, it is believed that this relates to the execution of RFC6121Section8_5_3_2_3_IqIntegrationTest

guusdk:
The RosterIntegrationTest frequently fails. Through debugging, we've established that this happens when the Roster's IQRequestHandler that processes roster pushes is not invoked (even though a corresponding stanza is received).

The source for this problem has been tracked down to RFC6121Section8_5_3_2_3_IqIntegrationTest. This test replaces IQRequestHandlers with an instance that's used by the test. The intention is to restore the handler that was replaced after the test has finished execution.

This commit fixes a bug that prevented the restoration of the original IQRequestHandler.

In the test, the handlers for multiple connections are replaced. The old values, however, are not stored _per connection_. This causes some connections to not get their original handler restored.

### DIFF
--- a/src/main/java/org/igniterealtime/smack/inttest/rfc6121/section8/RFC6121Section8_5_3_2_3_IqIntegrationTest.java
+++ b/src/main/java/org/igniterealtime/smack/inttest/rfc6121/section8/RFC6121Section8_5_3_2_3_IqIntegrationTest.java
@@ -453,7 +453,7 @@ public class RFC6121Section8_5_3_2_3_IqIntegrationTest extends AbstractSmackInte
                     break;
             }
 
-            final Map<IQRequestHandler, IQRequestHandler> receivedHandlers = new HashMap<>();
+            final Map<EntityFullJid, Map<IQRequestHandler, IQRequestHandler>> receivedHandlers = new HashMap<>();
             try {
                 // Setup test fixture: create connections for the additional resources (based on the user used for 'conTwo').
                 for (final AbstractXMPPConnection additionalConnection : additionalConnections) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/igniterealtime/smack/inttest/rfc6121/section8/RFC6121Section8_5_3_2_3_IqIntegrationTest.java
```diff
@@ -486,7 +486,13 @@ public IQ handleIQRequest(IQ iq) {
                         }
                     };
                     final IQRequestHandler oldHandler = resourceConnection.registerIQRequestHandler(needleDetector);
-                    receivedHandlers.put(needleDetector, oldHandler); // keep track so that the handler can be removed again.
+
+                    // keep track so that the handler can be removed again.
+                    final Map<IQRequestHandler, IQRequestHandler> receivedHandler = new HashMap<>();
+                    if (receivedHandler.put(needleDetector, oldHandler) != null) {
+                        throw new IllegalStateException("Bug in code: unexpectedly have more than one IQ handler the same detector for the same connection.");
```
Fishbowler, line 493: Nitpick
```suggestion
                        throw new IllegalStateException("Bug in code: unexpectedly have more than one IQ handler with the same detector for the same connection.");
```

Fishbowler (approved):

PR #113, opened 2025-08-23
Ensure that test users leave MUC room

### LINKED ISSUE #112: Inconsistent test failures for some test that relate to MUC
Some tests that relate to MUC functionality frequently (but not consistently) fail.

Examples:
```
MultiUserChatIntegrationTest.XEP0045/MultiUserChatIntegrationTest.mucDestroyTestOccupant (Normal): output/MultiUserChatIntegrationTest.mucDestroyOwnerTest (Normal).log#L1
Expected smack-inttest-two-khnsb@example.org/two-khnsb to no longer be in any rooms after smack-inttest-destroy-occupant-khnsb-2jhhg9@conference.example.org was destroyed. But it is still in [smack-inttest-xep0433-sort-khnsb-12@conference.example.org, smack-inttest-xep0433-sort-khnsb-14@conference.example.org, smack-inttest-xep0433-sort-khnsb-16@conference.example.org, smack-inttest-xep0433-sort-khnsb-18@conference.example.org, smack-inttest-xep0433-sort-khnsb-20@conference.example.org, smack-inttest-xep0433-sort-khnsb-22@conference.example.org, smack-inttest-xep0433-sort-khnsb-24@conference.example.org] ==> expected: <0> but was: <7>
```

```
MultiUserChatIntegrationTest.XEP0045/MultiUserChatIntegrationTest.mucDestroyOwnerTest (Normal): output/MultiUserChatIntegrationTest.mucDestroyOwnerTest (Normal).log#L1
Expected smack-inttest-one-khnsb@example.org/one-khnsb to no longer be in any rooms after smack-inttest-destroy-owner-khnsb-6rk1t6@conference.example.org was destroyed. But it is still in [smack-inttest-xep0433-sort-khnsb-11@conference.example.org, smack-inttest-xep0433-sort-khnsb-12@conference.example.org, smack-inttest-xep0433-sort-khnsb-13@conference.example.org, smack-inttest-xep0433-sort-khnsb-14@conference.example.org, smack-inttest-xep0433-sort-khnsb-15@conference.example.org, smack-inttest-xep0433-sort-khnsb-16@conference.example.org, smack-inttest-xep0433-sort-khnsb-17@conference.example.org, smack-inttest-xep0433-sort-khnsb-18@conference.example.org, smack-inttest-xep0433-sort-khnsb-19@conference.example.org, smack-inttest-xep0433-sort-khnsb-20@conference.example.org, smack-inttest-xep0433-sort-khnsb-21@conference.example.org, smack-inttest-xep0433-sort-khnsb-22@conference.example.org, smack-inttest-xep0433-sort-khnsb-23@conference.example.org, smack-inttest-xep0433-sort-khnsb-24@conference.example.org, smack-inttest-xep0433-sort-khnsb-25@conference.example.org] ==> expected: <0> but was: <15>
```

It seems that there's a test (XEP-0433 related) that lets test users join, but not leave, MUC rooms. This then affects other tests.

The inconsistency is probably caused by the tests running in a random order. This problem would occur only when the XEP-0433 test runs before the failing tests.

guusdk:
When test users join but not leave a MUC room during a test, this may affect other tests.

Each test should leave the test fixture 'clean': undo any changes that were made. This prevents tests affecting eachother.

This commit ensures that users leave MUCs when the test is over, for the XEP-0433 related test.

fixes #112

### DIFF
--- a/src/main/java/org/igniterealtime/smack/inttest/xep0433/ExtendedChannelSearchSortIntegrationTest.java
+++ b/src/main/java/org/igniterealtime/smack/inttest/xep0433/ExtendedChannelSearchSortIntegrationTest.java
@@ -28,6 +28,7 @@ import org.jivesoftware.smack.packet.IQ;
 import org.jivesoftware.smack.packet.StanzaError;
 import org.jivesoftware.smack.provider.ProviderManager;
 import org.jivesoftware.smackx.disco.ServiceDiscoveryManager;
+import org.jivesoftware.smackx.muc.MultiUserChat;
 import org.jivesoftware.smackx.muc.MultiUserChatManager;
 import org.jivesoftware.smackx.xdata.FormField;
 import org.jivesoftware.smackx.xdata.form.FillableForm;
@@ -40,8 +41,11 @@ import org.jxmpp.jid.parts.Localpart;
 import org.jxmpp.jid.parts.Resourcepart;
 import org.jxmpp.stringprep.XmppStringprepException;
 
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Set;
+import java.util.logging.Level;
 import java.util.stream.Collectors;
 
 import static org.junit.jupiter.api.Assertions.*;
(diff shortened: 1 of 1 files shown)


guusdk:
I've been toying with ideas like that, but for now, I've not been able to come up with a solution that doesn't have major drawbacks:
- We could try reconnecting each connection after each test. I suspect that this might make some connection provisioning mechanisms that we currently support harder (or impossible) to use. It would likely add considerably to the amount of data being logged, as well as the runtime execution time. If it's performed only as a reconnect on the same XmppConnection object, registered listeners and whatnot may still be present.
- In beta6, Smack introduces the concept of having auto-closable handlers (with `try-with-resource` constructs). That's certainly handy, but applicable only in scenario's where a handler is being created that can reasonably be expected to be discarded immediately after use. I don't think this should apply to MUCs and Rosters, as most of those objects (and likely, their event listeners) would, with general usage of Smack, be considered to be long-living things.

What we could do is improve our detection of tests that have implementation issues. One idea that I had for that is to execute the tests twice, in reverse order. If there's a difference in the collection of tests that fail in each test run, that's a strong indication that tests, somehow, affect each-other. I've captured this idea in this ticket: https://github.com/XMPP-Interop-Testing/smack-sint-server-extensions/issues/114

I'd love to have more tools in our belt, but I can't immediately think of others.

PR #121, opened 2025-09-03
Improve message error response detection

### LINKED ISSUE #120: Avoid unsafe correlation between message stanza and an error response
When a message stanza is sent and an error response is returned, it cannot be assumed that the returned stanza:
- contains the same message body as in the request
- uses a stanza ID attribute

guusdk:
Sadly, we can't depend on much characteristics to determine if a stanza is an error response to an earlier message stanza:
- IDs aren't required to be copied
- the body text isn't always copied (looking at Prosody)

This commit relaxes how those error stanzas are detected to just being a stanza of type error. That's rather loose, which may trip false positivies. It no longer fails to identify errors that are received.

fixes #120

Additionally, some surplus brackets were removed when used in lambdas.


Flowdalic:
> * IDs aren't required to be copied

I think it is required that the ID is preserved: RFC 6120 says:

> If the generated stanza was <message/> or <presence/> and
       included an 'id' attribute then it is REQUIRED for the error
       stanza to also include an 'id' attribute.

Fishbowler (approved):

PR #122, opened 2025-09-04
New tests for XEP-0060 Publish Subscribe, section 5

guusdk:


src/main/java/org/igniterealtime/smack/inttest/xep0060/PubSubUtils.java
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    /**
+     * Asserts that the provided service discovery item is a representation of a pub/sub node with a particular ID,
+     * using any of the addressing mechanisms defined in section 6.8 of XEP-0060.
+     *
+     * @param expectedNodeId The ID of the node that is expected to be represented by the item
+     * @param item a service discovery item
+     * @param message error message used when the assertion fails.
+     * @see <a href="https://xmpp.org/extensions/xep-0060.html#addressing">XEP-0060 Publish-Subscribe, Section 6.8 Addressing</a>
```
Fishbowler, line 70: Fix numbering

```suggestion
     * Asserts that the provided service discovery item is a representation of a pub/sub node with a particular ID,
     * using any of the addressing mechanisms defined in section 4.6 of XEP-0060.
     *
     * @param expectedNodeId The ID of the node that is expected to be represented by the item
     * @param item a service discovery item
     * @param message error message used when the assertion fails.
     * @see <a href="https://xmpp.org/extensions/xep-0060.html#addressing">XEP-0060 Publish-Subscribe, Section 4.6 Addressing</a>
```

src/main/java/org/igniterealtime/smack/inttest/xep0060/PubSubUtils.java
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    /**
+     * Asserts that the provided service discovery item is a representation of a pub/sub node with a particular ID,
+     * using the 'JID addressing' mechanism, as described in section 6.8.1 of XEP-0060.
+     *
+     * @param expectedNodeId The ID of the node that is expected to be represented by the item
+     * @param item a service discovery item
+     * @param message error message used when the assertion fails.
+     * @see <a href="https://xmpp.org/extensions/xep-0060.html#addressing-jid">XEP-0060 Publish-Subscribe, Section 6.8.1 JID</a>
```
Fishbowler, line 95: Fix numbering

```suggestion
     * Asserts that the provided service discovery item is a representation of a pub/sub node with a particular ID,
     * using the 'JID addressing' mechanism, as described in section 4.6.1 of XEP-0060.
     *
     * @param expectedNodeId The ID of the node that is expected to be represented by the item
     * @param item a service discovery item
     * @param message error message used when the assertion fails.
     * @see <a href="https://xmpp.org/extensions/xep-0060.html#addressing-jid">XEP-0060 Publish-Subscribe, Section 4.6.1 JID</a>
```

src/main/java/org/igniterealtime/smack/inttest/xep0060/PubSubUtils.java
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    /**
+     * Asserts that the provided service discovery item is a representation of a pub/sub node with a particular ID,
+     * using the 'JID+NodeID addressing' mechanism, as described in section 6.8.2 of XEP-0060.
+     *
+     * @param expectedNodeId The ID of the node that is expected to be represented by the item
+     * @param item a service discovery item
+     * @param message error message used when the assertion fails.
+     * @see <a href="https://xmpp.org/extensions/xep-0060.html#addressing-jidnode">XEP-0060 Publish-Subscribe, Section 6.8.2 JID+NodeID</a>
```
Fishbowler, line 127: Fix numbering

```suggestion
     * Asserts that the provided service discovery item is a representation of a pub/sub node with a particular ID,
     * using the 'JID+NodeID addressing' mechanism, as described in section 4.6.2 of XEP-0060.
     *
     * @param expectedNodeId The ID of the node that is expected to be represented by the item
     * @param item a service discovery item
     * @param message error message used when the assertion fails.
     * @see <a href="https://xmpp.org/extensions/xep-0060.html#addressing-jidnode">XEP-0060 Publish-Subscribe, Section 4.6.2 JID+NodeID</a>
```

src/main/java/org/igniterealtime/smack/inttest/xep0060/section5/PubSubSection5_3_IntegrationTest.java
```diff
@@ -0,0 +1,258 @@
...
+
+/**
+ * Tests as defined in paragraph 5.3 "Discover Node Information" of section 5 "Entity Use Cases" of XEP-0060 "Publish-Subscribe".
+ *
+ * @author Guus der Kinderen, guus.der.kinderen@gmail.com
+ * @see <a href="https://xmpp.org/extensions/xep-0060.html#entity-nodes">XEP-0060: Publish-Subscribe</a>
+ */
+// TODO: Some of this (around 'Collection Nodes' and hierarchy) should be part for XEP-0248, not XEP-0060, although as of version 1.26.0 of XEP-0060 it is in that specification. Monitor later versions of the specification for changes (as suggested in https://mail.jabber.org/hyperkitty/list/standards@xmpp.org/thread/COEJQNNCEHHT2WFF46CWYYYVCL2NIOE4/ )
+@SpecificationReference(document = "XEP-0060", version = "1.26.0")
+public class PubSubSection5_3_IntegrationTest extends AbstractSmackIntegrationTest
```
Fishbowler, line 48: This file doesn't include an explicit test for this statement:

> Note: If a node has an identity type of "leaf", then it MUST NOT contain other nodes or collections (only items); if a node has an identity type of "collection", then it MUST NOT contain items (only other nodes or collections).

Do we get that as a guarantee from Smack?

Fishbowler (changes requested):
I've added fix suggestions for a few typos, and a question.

Interesting stuff in 5.6/5.7, where a server MAY support a thing. If it does, it MUST do this. If it doesn't, it MUST do that. Feels like a gap we could open an issue for resolving separately though?

src/main/java/org/igniterealtime/smack/inttest/xep0060/section5/PubSubSection5_3_IntegrationTest.java
guusdk: I'm not sure how we could test for that. We could generically assert that any result doesn't include mixed node types, but the chances of that happening on a system-under-test (which is probably an empty dev server), without us explicitly setting that up in the test fixture are negligible. At that point, I feel that this becomes a test for the bits of the XEP that specify how nodes are added (where, I assume, the 'MUST NOT contain other nodes or collections' statement is repeated).

PR #125, opened 2025-09-11
Skip ad-hoc tests when not provisioned with an admin account

guusdk:
If the runner is configured to use three explicitly named accounts (instead of an admin account that creates accounts as desired), the tests for ad-hoc commands crash. These tests require an administrative account.

In this comment, these tests raise a TestNotPossibleException, rather than cause a crash. This allows other tests to continue.

### DIFF
--- a/src/main/java/org/igniterealtime/smack/inttest/xep0133/AbstractAdHocCommandIntegrationTest.java
+++ b/src/main/java/org/igniterealtime/smack/inttest/xep0133/AbstractAdHocCommandIntegrationTest.java
@@ -52,10 +52,14 @@ public class AbstractAdHocCommandIntegrationTest extends AbstractSmackIntegratio
     public final AbstractXMPPConnection adminConnection;
     SmackIntegrationTestEnvironment environment;
 
-    public AbstractAdHocCommandIntegrationTest(SmackIntegrationTestEnvironment environment) throws SmackException, IOException, XMPPException, InterruptedException, InvocationTargetException, InstantiationException, IllegalAccessException {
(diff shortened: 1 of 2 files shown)


src/main/java/org/igniterealtime/smack/inttest/xep0133/AbstractAdHocCommandIntegrationTest.java
```diff
@@ -52,10 +52,14 @@ public class AbstractAdHocCommandIntegrationTest extends AbstractSmackIntegratio
...
     SmackIntegrationTestEnvironment environment;
 
-    public AbstractAdHocCommandIntegrationTest(SmackIntegrationTestEnvironment environment) throws SmackException, IOException, XMPPException, InterruptedException, InvocationTargetException, InstantiationException, IllegalAccessException {
+    public AbstractAdHocCommandIntegrationTest(SmackIntegrationTestEnvironment environment) throws SmackException, IOException, XMPPException, InterruptedException, InvocationTargetException, InstantiationException, IllegalAccessException, TestNotPossibleException
+    {
         super(environment);
         this.environment = environment;
 
+        if (sinttestConfiguration.adminAccountUsername == null) {
+            throw new TestNotPossibleException("This test requires an admin account to be configured.");
```
Fishbowler, line 61: How do you feel about pointing them at more docs? 

```suggestion
            throw new TestNotPossibleException("This test requires an admin account to be configured. Configuration instructions available at https://xmpp-interop-testing.github.io/");
```

Fishbowler (approved):

Issue #126, opened 2025-09-15
Add config options for test accounts


guusdk:
The tests should be configurable using three predefined accounts. The container images should allow those three accounts to be provided.

PR #133, opened 2025-09-21
CI: add tests for each account provisioning method

guusdk:
Prior to this PR, the workflows of this repository test the built java version, as well as one of the containers, against a running XMPP server.

In this PR, that test is expanded with new tests, that each use different test account provisioning options (#126)

As an aside, this PR also contains the following change: Replace boilerplate code with GitHub Action

The continuous integration workflow starts an Openfire server for testing. In this commit, all of the related instructions are replaced with the GitHub action that is provided in https://github.com/igniterealtime/launch-openfire-action which does the same thing.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,7 +7,7 @@ permissions:
 
 jobs:
   build-java:
-    name: Build
+    name: Build the test suite
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


guusdk:
Ideally, I'd like to split up the 'Build & Test the Docker image' job into several jobs (one that builds, and three that test). This requires more knowledge about how to pass around Docker images than that I possess. This isn't a hill for me to die on.

There's also a "Build with Podman" job that doesn't test things at all. Is this what builds the OCI file? Should that also be tested?

guusdk:
I think this all has to do with the JUnit testresult parser. Maybe we should just remove that in these workflows?

>  Do we want checks created?

I don't know what you mean by that.

Issue #134, opened 2025-09-21
Add tests for the docker container


Fishbowler:
#133 added tests for the Java launch of Admin/IBR/explicit account methods of user provisioning for the tests

Perhaps we need the same for Docker.

(We don't need the same for Podman/OCI, since that's using an identical entrypoint)

Fishbowler:
We discussed in the channel whether the BATS tests that we have for the entrypoint might guarantee the invocation of the JAR, and the JAR is already tested.

We still need _some_ test of the artifact we produce (which we do), but that doesn't mean that _this_ test needs to be at that level.

TBC...

PR #138, opened 2025-09-22
Add option to fail test run on impossible test

### LINKED ISSUE #137: Add config option for fail-on-impossible-test
Test may be impossible to run, for example, because the server that's being tested does not support a particular feature.

It is not unreasonable for users to assume that, upon a successful test execution result, all tests have passed. In the case of 'impossible' tests, this isn't necessarily the case: some test might not have executed at all.

Especially in scenarios where users have configured the test run to enable a specific subset of the tests (through configuration such as enabledTests and enabledSpecifications) it may be desirable to hard fail when a test was impossible to execute. This enforces that a test run was strictly successful in all tests it was configured to execute.

Since version 4.5.0-beta7, Smack supports a configuration option that allows a test run to be marked as 'failed' when one or more tests were impossible to execute (see https://github.com/igniterealtime/Smack/pull/663 for details).

This project should expose this configuration, so that it becomes usable for end-users.

guusdk:
The `failOnImpossibleTest` option that's added in this commit can fail the test run if any configured tests were impossible to execute, ensuring all intended tests actually ran.

fixes #137

### DIFF
--- a/test/entrypoint.bats
+++ b/test/entrypoint.bats
@@ -71,3 +71,10 @@ setup() {
   assert_output --partial "Running: java -Dsinttest"
   assert_output --partial "-Dsinttest.service=test.example"
 }
+
+@test "successfully applies 'failOnImpossibleTest'" {
+  run "$SCRIPT" --failOnImpossibleTest
+  assert_success
+  assert_output --partial "Running: java -Dsinttest"
+  assert_output --partial "-Dsinttest.failOnImpossibleTest=true"
+}


guusdk:
I'd like an explicit test that asserts that the container images support this option too. That probably requires #134 to be fixed first.

guusdk:
> Can I entice you to write a test in BATS to assert that the correct Java invocation is made?

I'm not sure if I understand the purpose or application of those tests - but I've given it a try. How's this?

